PR #2, opened 2016-10-09
Added test case for OMTG-DATAST

sushi2k:
I migrated the following test cases from the Google Doc:

OMTG-DATAST-001: Testing for Insecure Storage of Credentials and Keys 
OMTG-DATAST-002: Testing for Sensitive Data Disclosure in Log Files 
OMTG-DATAST-004: Testing for Sensitive Data Disclosure in Local Storage
OMTG-DATAST-005: Testing for Sensitive Data sent to 3rd Parties 
OMTG-DATAST-006: Test for Sensitive Data Disclosure in Process Memory

### DIFF
--- a/Document/Testcases/0x00_OMTG-DATAST.md
+++ b/Document/Testcases/0x00_OMTG-DATAST.md
@@ -19,7 +19,7 @@
 - OWASP MASVS: V2-1: "Verify that system credential storage facilities are used appropriately to store sensitive data, such as user credentials or cryptographic keys."
 - CWE: [Link to CWE issue]
 
-### OMTG-DATAST-009: OMTG-DATAST-009: Test for Sensitive Data in Backups
+### OMTG-DATAST-009: Test for Sensitive Data in Backups
 [General description]
 
 #### Detailed Guides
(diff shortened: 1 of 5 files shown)


b-mueller:
This is all content you wrote yourself right?

b-mueller:
"An app shouldn’t store any sensitive information like credentials, passwords or encryption keys (even when using security controls and mechanisms offered by the OS as a best practice to protect this information)."

I disagree with that. We can't state this and then go on to allow it in the requirements.

b-mueller:
"For black box testing, the memory should be analysed..." etc... we need to define some standards on how test cases are written. In the instruction sections we address the reader in the second person, e.g.:

* Analyze the memory bla bla bla...

I'll open an issue for this shortly!

sushi2k:
Yes, these are the things that I wrote myself.

There are definitely a lot of things that need to be aligned so the test cases are written in the same manner. 

> I disagree with that. We can't state this and then go on to allow it in the requirements.

For example the sentence you mentioned. This need to be aligned with the MASVS and this sentence is also definitely too hard and should be rewritten.

b-mueller:
By the way, the list of test cases is not the same as in the OMTG "Beta 2" - for example, "OMTG-DATAST-004 - Test for sensitive data disclosure in local storage" doesn't exist any more. There should be a 1:1 mapping of MASVS requirements to test cases. The original content from DATAST-004 needs to be merged into "OMTG-DATAST-001: Test Sensitive Data Storage" which maps to V2.1 in the MASVS.
Lets talk about it sometime next week and define all those requirements so we're all on the same page!

sushi2k:
Ok. Sorry, was not aware of that. Sure let's catch up next week, then I can update it accordingly.

PR #35, opened 2016-11-26
Update 0x02-Frontispiece.md

milanowasp:


b-mueller:
@milanowasp, as discussed I added a revision history and full list of original authors, as well as special credit to you for initiating the guide.

PR #36, opened 2016-11-26
Adding Foreword

milanowasp:


b-mueller:
@milanowasp this is a copy of Jeff William's foreword from the OWASP Testing guide 3.0. I'm going to add original content here.

PR #54, opened 2017-01-07
Adding OMTG-NET-002: Test X.509 certificate verification test

xep624:


b-mueller:
Thanks for the contribution Pawel!
It appears you have committed some IntelliJ XML files as well, please add them to your .gitignore and re-commit.
The white-box testing part is great, however for the black-box testing part we need more generic instructions (not dependent on a specific scanner such as Mallodroid). Such as: Either decompile and do the same as in white-box testing, or dynamic analysis using a proxy (change the certificate an verify that the network connection fails).

xep624:
Thanks for your feedback Bernhard! not sure how those xml files appeared there, but let me fix this issue. I'll also add more instructions to black box testing

PR #59, opened 2017-01-11
Added Method to recover ipa on non-jailbroken devices

ryantzj:
As title


ryantzj:
was pondering the same thing earlier but could not find any good solution to it, i have added a TODO on extracting ipa from ios devices.

PR #66, opened 2017-01-27
Update 0x04a-Android.md

luander:
I added this section because it seems fit in this section. Feel free to comment about it or suggest to move to another place.

### DIFF
--- a/Document/0x04a-Android.md
+++ b/Document/0x04a-Android.md
@@ -189,119 +189,61 @@ drwxr-xr-x    9 sven  staff   306B Dec  5 16:29 smali
 * **res**: A directory containing resources not compiled into resources.arsc.
 * **smali**: A directory containing the disassembled Dalvik Bytecode in Smali. Smali is a human readable representation of the Dalvik executable.
 
+
 ### Android Users and Groups
 
-Android is a system based on Linux, however it does not deal with users the same way Linux does. It does not have a /etc/password file describing a list of Linux users in the system. Instead Android contains a fixed set of users and groups and they are used to isolate processes and grant permissions.
+Android is a system based on Linux, however it does not deal with users the same way Linux does. It does not have a _/etc/password_ file describing a list of Linux users in the system. Instead Android contains a fixed set of users and groups and they are used to isolate processes and grant permissions.
 File [system/core/include/private/android_filesystem_config.h](http://androidxref.com/7.1.1_r6/xref/system/core/include/private/android_filesystem_config.h) shows the complete list of the predefined users and groups mapped to numbers.
-File below depicts all users defined for Android Nougat:
+File below depicts some of the users defined for Android Nougat:
 ```
-    /* This is the master Users and Groups config for the platform.
-     * DO NOT EVER RENUMBER
-     */
-
     #define AID_ROOT             0  /* traditional unix root user */
 
     #define AID_SYSTEM        1000  /* system server */
-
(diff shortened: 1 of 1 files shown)


Document/0x04a-Android.md
```diff
@@ -189,6 +189,121 @@ drwxr-xr-x    9 sven  staff   306B Dec  5 16:29 smali
 * **res**: A directory containing resources not compiled into resources.arsc.
 * **smali**: A directory containing the disassembled Dalvik Bytecode in Smali. Smali is a human readable representation of the Dalvik executable.
 
+### Android Users and Groups
+
+Android is a system based on Linux, however it does not deal with users the same way Linux does. It does not have a /etc/password file describing a list of Linux users in the system. Instead Android contains a fixed set of users and groups and they are used to isolate processes and grant permissions.
+File [system/core/include/private/android_filesystem_config.h](http://androidxref.com/7.1.1_r6/xref/system/core/include/private/android_filesystem_config.h) shows the complete list of the predefined users and groups mapped to numbers.
+File below depicts all users defined for Android Nougat:
```
sushi2k: Can you please reduce the list to two or three users to show that subsystems in linux have their own users to enforce sandboxing? No need to copy the whole list.

Document/0x04a-Android.md
```diff
@@ -189,6 +189,121 @@ drwxr-xr-x    9 sven  staff   306B Dec  5 16:29 smali
...
+
+    #define AID_SHARED_GID_START 50000 /* start of gids for apps in each user to share */
+    #define AID_SHARED_GID_END   59999 /* start of gids for apps in each user to share */
+```
+
+What is shown when the `id` command is executed on an Android shell is the user and group ids applied for the current process.
+```
+$ adb shell id
+uid=2000(shell) gid=2000(shell) groups=2000(shell),1004(input),1007(log),1011(adb),1015(sdcard_rw),1028(sdcard_r),3001(net_bt_admin),3002(net_bt),3003(inet),3006(net_bw_stats),3009(readproc) context=u:r:shell:s0
+```
```
sushi2k: Can you also add a paragraph that for each installed App a new user will be created and only this users has access to the data directory of the app etc?

sushi2k:
Hi @luander,

thanks for contributing. The user section makes sense indeed. Can you please address the two comments? Then I will merge it. 

Cheers,

Sven

luander:
Hi @sushi2k,
Thank you for your suggestions, I agree with them all. 
Please review my changes and suggest new improvements.

Cheers.

Document/0x04a-Android.md
```diff
@@ -189,6 +189,63 @@ drwxr-xr-x    9 sven  staff   306B Dec  5 16:29 smali
...
+File below depicts some of the users defined for Android Nougat:
+```
+    #define AID_ROOT             0  /* traditional unix root user */
+
+    #define AID_SYSTEM        1000  /* system server */
+	...
+    #define AID_SHELL         2000  /* adb and debug shell user */
+	...
+    #define AID_APP          10000  /* first app user */
+	...
```
sushi2k: Can you please add references if you are using content from other sources? For example it seems you used this as reference for this section https://github.com/keesj/gomo/wiki/AndroidSecurityUserAndGroups.

So put in the link as reference at the end of the section. See this for further details on how to use external references: https://github.com/OWASP/owasp-mstg/blob/master/style_guide.md#external-references

sushi2k:
Looks good, but please put in a reference.

luander:
References added.

PR #78, opened 2017-02-10
Update the headings for net

commjoen:
Here is an update for the headings for the net chapters. Hope to fill in lots of the content in the upcoming month


b-mueller:
Thanks @commjoen!

PR #95, opened 2017-02-17
Add elements on signing and publishing

romualdszkudlarek:
Also removed todo list at the beginning of the chapter as it becomes obsolete.


b-mueller:
Thanks @romualdszkudlarek

PR #113, opened 2017-02-23
Fixed typos.

demonbensa:
On line 386 it reads "Broadcast Receivers is that they be affected a priority". This sentence does not make sense and should be revised.


romualdszkudlarek:
Thanks for the updates @demonbensa ! It seems you have proposed some fixes in your pull request and I agree with all of them (I am the author of these lines), but haven't fixed the one mentioned in the headline ("Broadcast Receivers is that they be affected a priority"). I will fix it right now.

PR #114, opened 2017-02-23
0x5a - fixed typo following @demonbensa feedback

romualdszkudlarek:
Clarified sentence "Broadcast Receivers is that they be affected a priority"


demonbensa:
Thank you for considering my recommendations.

Now that you have made a change, I think I understand what you are trying to say and suggest the following version:

Broadcast Receivers is that they can be affected **by** priority.

b-mueller:
Wouldn't it be best to say "assigned a priority"? :)

romualdszkudlarek:
"assigned a priority" seems best to me. Thanks to both of you!

PR #117, opened 2017-02-24
0x5b - Add suggestions on additional elements

romualdszkudlarek:
Hardware / Rooting your device : I think it can be beneficial to OWASP reputation and legal safety to add some clarifications on the consequences of rooting.


b-mueller:
Good idea. I'd never have thought of that :)

PR #131, opened 2017-02-25
0x4 - Clarified statement against tampering

romualdszkudlarek:


romualdszkudlarek:
When typing "tampering defenses" in Google, I get a majority of answers concerning "anti-tampering defenses", which suggests "anti-tampering" is used in most cases. Is there a linguist around? ;-)

If you've used that statement before ("tampering defenses") and if no one complained, then your version is fine for me. Feel free to close the pull request if relevant.

b-mueller:
haha yeah, we should get a linguist to review the whole thing at the end :)

b-mueller:
You're right, Google results say "anti-tampering defenses". I'll change it.

PR #136, opened 2017-02-25
0x4 - Fixed typo "to contain of"

romualdszkudlarek:


romualdszkudlarek:
Let's change it to "consist" then!

PR #144, opened 2017-02-25
0x4 - Clarified statement

romualdszkudlarek:
"security of a program" becomes "security level of a program"


b-mueller:
Hey @romualdszkudlarek, I indeed meant to say "prove security" here (its and academic-type expression).

PR #153, opened 2017-02-26
0x5b - Added a section on risks of rooting a device

romualdszkudlarek:


b-mueller:
Thank you @romualdszkudlarek. I would change it a bit stylistically and address security testers instead of a general audience  - presumably they know what rooting means, but we should describe the technical details a bit more, and why it is necessary to root a device for testing apps (e.g, "for security testing purposes you'll most likely want to  root your Android device... etc.).

romualdszkudlarek:
Thanks for the feedback @b-mueller , you are right, I will rephrase for a more technical audience. However, I do not want to add too many technical details, as my original purpose was to have a disclaimer only (to protect OWASP on the legal side). I agree technical details have to be added in 0x5b (how to root, why rooting, ...), but I think it should be in another section (cf 3rd section of the 'Rooting your device' paragraph : -- TODO : Boot Process Description -- -- TODO : Boot Loaders and ROMs--). I will propose another disclaimer. What do you think?

b-mueller:
Yeah, technical details should be further down. Keep the disclaimer paragraph short then I would say?

romualdszkudlarek:
Let's make it this way. I'll close this pull request and will commit another version.

PR #170, opened 2017-03-01
0x5b - Updated proposal for a disclaimer on rooting a device

romualdszkudlarek:


b-mueller:
Very nice, thanks @romualdszkudlarek

PR #175, opened 2017-03-02
Changes secure software testing

prabhant:
secure software testing was a very vague term
either it should be software testing or security testing in SDLC


prabhant:
suggested changes done.

PR #176, opened 2017-03-05
Update README.md

romantic668:
Hi, I am doing a technical writing project. I think it is important to keep the content concise and to the point. Hope it helps


b-mueller:
Does make it more concise, thanks!

PR #183, opened 2017-03-07
Update 0x06b-Basic-Security-Testing.md

skosowski:
Update


skosowski:
I removed info on checking binary protections and amended one line or so.
Instruction regarding dumpdecrypted was unnecessarily complicated. Note that not all people have XCode to compile the project. Clutch is easy to use and available directly from the Cydia store.

PR #235, opened 2017-03-23
Bactis/newemulatorcontent

bactis:
added some basic static emulator checks...let me know if formatting is ok before i add all the dynamic ones.


b-mueller:
Thank @bactis

PR #416, opened 2017-04-10
Frida solution for uncrackable1 crackme

mhelwig:
A draft for a Frida based solution to the Android Uncrackable1 crackme.


Document/0x05c-Reverse-Engineering-and-Tampering.md
```diff
@@ -697,6 +697,217 @@ Java.perform(function () {
...
+            alertDialog.setTitle((CharSequence)"Nope...");
+            alertDialog.setMessage((CharSequence)"That's not it. Try again.");
+        }
+        alertDialog.setButton(-3, (CharSequence)"OK", (DialogInterface.OnClickListener)new c(this));
+        alertDialog.show();
+    }
+}
+```
+
+Notice the `Root detetected` message in the `onCreate` method and the various methods called in the the `if`-statement before which perform the actual root checks. Also note the `This is unacceptable...` message from the first method of the class, `private void a`. Obviously, this is where the dialog box gets displayed. There is a `alertDialog.onClickListener`callback set in the `setButton` method call which is responsible for closing the application via `System.exit(0)` after successful root detection. Using Frida, we can prevent the app from exiting by hooking the callback.
```
anantshri: typo: detected. 

It would be better if you can reference by line number.

mhelwig: I don't see line numbers on other codelistings here, won't add them manually because that doesn't make sense. Line numbers were great if could add them globally.

PR #426, opened 2017-04-21
Update 0x04-Testing-Processes-and-Techniques.md

bengardiner-at-irdeto:
There are static analysis tools that do not require source code.

I think the authors already recognize this fact as there is an inconsistency here as well: the leading paragraph of the *Static Analysis* section says `the developers must provide the source code or compiled IPA/APK files of the mobile application for programmatic analysis`. Whereas the passage for which I am proposing a change currently states that source code will be required.


b-mueller:
Thanks for the fix Ben.

PR #441, opened 2017-05-08
Update 0x05f-Testing-Authentication.md

L3Osi13nT:
A small change on references under Testing the Password Policy.


sushi2k:
Why not link directly to the Web App Testing Guide password policy? Think that might be the best 

https://www.owasp.org/index.php/Testing_for_Weak_password_policy_(OTG-AUTHN-007)

L3Osi13nT:
Yes. Added.

PR #442, opened 2017-05-08
Update 0x05f-Testing-Authentication.md

L3Osi13nT:
Changes on Static Analysis and Remediation under Testing the Password Policy.

### DIFF
--- a/Document/0x05f-Testing-Authentication.md
+++ b/Document/0x05f-Testing-Authentication.md
@@ -155,7 +155,27 @@ Many Apps do not automatically logout a user, because of customer convenience. T
 
 #### Overview
 
-Password policy is a special countermeasure to avoid any application user on creating passwords that are easily guessable. 
+Password strength is a key concern when using passwords for authentication. Password policy defines controls that end users should adhere to. Password Length, Password Complexity and Password Topologies should properly be included in the Password Policy. A "strong" password policy makes it difficult or even improbable for one to guess the password through either manual or automated means. 
+
+A good password policy should defines following controls in order to avoid password guessing attacks or even brute-forcing. 
+#####  Password Length
+* Minimum length of the passwords should be enforced by the application.
+* Maximum password length should not be set too low, as it will prevent users from creating passphrases. Typical maximum length is 128 characters.
+
+##### Password Complexity
+* Password must meet at least 3 out of the following 4 complexity rules
+1. at least 1 uppercase character (A-Z)
+2. at least 1 lowercase character (a-z)
+3. at least 1 digit (0-9)
+4. at least 1 special character (punctuation)
+* at least 10 characters
+* at most 128 characters
+* not more than 2 identical characters in a row
+
+##### Password Topologies
+* Ban commonly used password topologies.
+* Force multiple users to use different password topologies.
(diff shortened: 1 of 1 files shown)


Document/0x05f-Testing-Authentication.md
```diff
@@ -155,10 +155,33 @@ Many Apps do not automatically logout a user, because of customer convenience. T
...
+* NonAlphanumericCharacterRule - Does a password contain a non-alphanumeric character
+* NumericalSequenceRule - Does a password contain a numerical sequence
+* RegexRule - Does a password match a regular expression
+* RepeatCharacterRegexRule - Does a password contain a repeated character
+* SequenceRule - Does a password contain a keyboard sequence
+* SourceRule - Does a password match the password from another system or source
+* QwertySequenceRule - Does a password contain a QWERTY keyboard sequence
+* UppercaseCharacterRule - Does a password contain an uppercase character
+* UsernameRule - Does a password contain a username
+* WhitespaceRule - Does a password contain whitespace
```
sushi2k: I would remove this listing, as I think a password policy should allow any character as stated here, this should also be used as a reference:
https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Implement_Proper_Password_Strength_Controls 

Static Analysis of the password policy would involve server code and you could list down an example for a regex that might be used to check for a strong password. You could describe how you can check if the regex is correct through something like http://regexr.com/ and see what passwords are matching the regex

Document/0x05f-Testing-Authentication.md
```diff
@@ -191,8 +224,8 @@ Many Apps do not automatically logout a user, because of customer convenience. T
 
 ##### Info
 
-- [1] Meyer's Recipe for Tomato Soup - http://www.finecooking.com/recipes/meyers-classic-tomato-soup.aspx
-- [2] Another Informational Article - http://www.securityfans.com/informational_article.html
+- [1] VT Password - https://code.google.com/archive/p/vt-middleware/wikis/vtpassword.wiki
```
sushi2k: I would remove this, as this is from 2013 and not maintained since

L3Osi13nT:
Will do the changes. thanks for reviewing.

sushi2k:
Hi @L3Osi13nT. When do you think you can make an update?

Document/0x05f-Testing-Authentication.md
```diff
@@ -155,10 +155,53 @@ Many Apps do not automatically logout a user, because of customer convenience. T
 
 #### Overview
 
+Password strength is a key concern when using passwords for authentication. Password policy defines controls that end users should adhere to. Password Length, Password Complexity and Password Topologies should properly be included in the Password Policy. A "strong" password policy makes it difficult or even improbable for one to guess the password through either manual or automated means.
```
Sjord: > defines controls
I am not sure "controls" is the right word here. Maybe "requirements"?

> Password Length, Password Complexity and Password Topologies should properly be included in the Password Policy.
I don't think all these words should start with capitals.

> improbable
Maybe "infeasible" is a better word?

sushi2k: Will change that once merged. Thanks @Sjord

PR #448, opened 2017-05-10
Debugging with IDE 

oguzhantopgul:
I added some content on debugging android applications from decompiled sources using Android Studio. I debugged Uncrackable1 app as an example.
Please review, proofread and let me know if any corrections needed


b-mueller:
Fantastic work @oguzhantopgul! I don't see any issues so far so I'll merge the PR, and do another round of proofreading later.

PR #449, opened 2017-05-12
0x5j/device binding

commjoen:
Hi all,
this is a first pull-request on device-binding. It took a little longer because of the changes in Android O. The write-up is not yet complete (references need to be added at the end). Please review it with care so that I can finalize it.

Grz,
Jeroen


b-mueller:
Fantastic write-up! Will merge it as-is and perhaps do a few cosmetic changes later on.

PR #456, opened 2017-05-16
Difference of encryption modes

EhsanMashhadi:
Difference of encryption modes and weak ECB mode


EhsanMashhadi:
@b-mueller Sure i will do changes

Sjord:
Doing encryption correctly can fill another whole book. Up to what extent is encryption in scope for the MSTG?

Document/0x05e-Testing-Cryptography.md
```diff
@@ -216,6 +216,14 @@ Example initialization of DES algorithm, that is considered weak:
 ```Java
 Cipher cipher = Cipher.getInstance("DES");
 ```
+##### Block cipher encryption modes
+Do not use ECB encryption mode, it is basically raw cipher. For each block of input, you encrypt that block and get some output. The problem with this transform is that any resident properties of the plaintext might well show up in the ciphertext – possibly not as clearly – that's what blocks and key schedules are supposed to protect againt, but analyzing the patterns you may be able to deduce properties that you otherwise thought were hidden.
```
Sjord: Nice addition of the dangers of ECB mode. I think this could be more clear if you clearly separate sentences. For example,

> Do not use ECB encryption mode, it is basically raw cipher.

These are actually two sentences that should be separated by a period:

> Do not use ECB encryption mode. ECB mode is basically raw cipher.

Or they should be joined by indicating the relation:

> Since ECB encryption mode is basically raw cipher, it should not be used.

commjoen: It might be nice to augment the warning with a nice set of pictures used in Wikipedia ? (https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation)

b-mueller:
> Doing encryption correctly can fill another whole book. Up to what extent is encryption in scope for the MSTG?

@Sjord we should cover the basics of everything that's relevant in the mobile world, both for encrypting data locally and encrypting network traffic. Plus, recommended settings, such as min. key length etc.

I'm not so happy with how this is currently structured. We're now moving general best practices into its own chapter, which will also include the crypto stuff. We'll also need to think about what scenarios are best practice for local storage. For example, we say that one shouldn't use hardcoded keys, and later on we then talk about encrypting data with AES. It's inconsistent.

b-mueller:
@sushi2k this is part of what needs to be in 0x4b...

sushi2k:
@EhsanMashhadi Just merged your request. Will also add the diagram mentioned by @luizlohn

PR #474, opened 2017-05-23
Added Write-up on Different Types of Mobile Apps

kongwenbin:
And also modified the URLs to make them suitable for "printing" as git book


b-mueller:
Very cool! There's a few grammar bugs in there, but we'll fix those later.

PR #481, opened 2017-05-24
[android/backup] Auto Backup is enabled by default

andreashappe:
..when using Android 6.0 onwards. The existing paragraph was framed as if the opposite would be the case.


sushi2k:
Thanks @andreashappe and @pmilosev.

PR #484, opened 2017-05-25
Review: basic security testing on android

pmilosev:


Document/0x05b-Basic-Security_Testing.md
```diff
@@ -153,11 +153,10 @@ See also test case "Testing Custom Certificate Stores and SSL Pinning" for furth
 
 ##### Root Detection
 
-To implement root detection on Android, libraries like RootBeer<sup>[14]</sup> or custom checks are used to verify if the device is rooted or not. See also test case "Testing Root Detection" and "Testing Advanced Root Detection" for further details.
+To implement root detection on Android, libraries like RootBeer<sup>[14]</sup> or custom checks are used to verify if the device is rooted or not. Please note that root detection mechanisms are relatively easy to circumvent and should only be used to support, but not base the app security upon. See also test case "Testing Root Detection" and "Testing Advanced Root Detection" for further details.
```
Sjord: > To implement root detection on Android, libraries like RootBeer<sup>[14]</sup> or custom checks are used to verify if the device is rooted or not.

This sentence contains two times a subsentence with "to": "To implement root detection on Android", "to verify if the device is rooted or not". This makes it a bit awkward, and it still does not explain why you would want root protection. Maybe you can change this sentence to make it match your added sentence better.

> Please note that

I am not sure this properly indicates the relation to the rest of the paragraph.

> root detection mechanisms are relatively easy to circumvent

Maybe you can provide a pointer about how the mechanisms can be bypassed?

> be used to support, but not base the app security upon.

I think this part of the sentence doesn't really work, because in the first part of the sentence it is not clear what should be supported, and the "upon" messes that up. 

I think the fact that root detection is easy to bypass is useful information for a mobile pentester, but I am not sure whether it belongs in this chapter. This chapter is more about getting an app working on a rooted device so that you can test it, and not about the design decisions when creating a mobile app. Even so, this chapter could use some pointers about how to bypass the root detection, or at least a note that it is possible.

Document/0x05b-Basic-Security_Testing.md
```diff
@@ -8,7 +8,7 @@ This section will give an overview of different methods on how an Android app ca
 
 #### Preparation
 
-The goal of a test is to verify if the app and the endpoint(s) it's communicating with, are implemented in a secure way. Several security controls like SSL Pinning or root detection might be implemented. These can slow down the testing dramatically and might already take days to bypass, depending on the implementation.
+The goal of a test is to verify if the app and the client side of it's communication with the endpoint(s), are implemented in a secure way. Several security controls like SSL Pinning or root detection might be implemented. These can slow down the testing dramatically and might already take days to bypass, depending on the implementation.
```
b-mueller: > The goal of a test is to verify if the app and the endpoint(s) it's communicating with, are implemented in a secure way. 

it's -> its, and remove the comma after "with"

also "verify if the app and the client side of its communication with the endpoint(s)" is a weird-sounding construction. I'd rather write "verify if network communication is implemented securely on the client".

b-mueller:
Merging this pull request, will address @Sjord's feedback as well.

pmilosev:
Thanks for the correction and the merge.
I had few escalations on other projects these days and could not update the texts.

PR #490, opened 2017-05-30
Added Write-up on Testing Random Number Generation

kongwenbin:
Also fixed some typo and add some comments


b-mueller:
Thanks! Will merge and follow up with some grammar corrections.

PR #494, opened 2017-05-31
Added content for Sensitive Data in Backups and Keyboard Cache

kongwenbin:


Sjord:
Maybe this is a useful resource in this context: [How do I prevent files from being backed up to iCloud and iTunes?](https://developer.apple.com/library/content/qa/qa1719/_index.html)

b-mueller:
Thanks for adding the content re. backups Wen Bin! There is a separate test case for this though in the same file: "Testing for Sensitive Data in Backups". Can you move the info on backups to that other test case.

kongwenbin:
I checked, the content were added to the "Testing for Sensitive Data in Backups" section. Do you mean another section?

b-mueller:
@kongwenbin oops you're right, I mis-saw that :)

PR #495, opened 2017-06-01
Update Cryptography

andreashappe:
* better introducion
* restructure flow/order, now this should feel better
* distribute text between existing test cases
* add description for key derivation functions
* add more reference links


b-mueller:
Awesome work, thanks!

PR #515, opened 2017-06-13
JWT Test Case

sushi2k:


Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,98 @@ If any of these two conditions raise an issue, reject the request and do not all
 
 
 
+### Testing JSON Web Token (JWT)
+
+#### Overview
+
+JSON Web Token (JWT) ensures the integrity and secure transmission of information within a JSON object between two parties and is defined in RFC 7519<sup>[1]</sup>. For mobile apps it's more and more used to authenticate both, the message sender and receiver.
```
Sjord: > JSON Web Token (JWT) ensures the integrity and secure transmission

It doesn't really ensure the "secure transmission" in the sense that it does not provide confidentiality.

> For mobile apps it's more and more used to authenticate both, the message sender and receiver.

I think this is better without the comma: "both the message sender and receiver".

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,98 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+  base64UrlEncode(payload),
+  secret
+)
+```
+
+JWT implementations are available for all major programming languages, like PHP<sup>[2]</sup> or Java Spring<sup>[3]</sup>.
+
+#### Static Analysis
+
+Identify the JWT library that is used on server and client side. It should be checked if there are any known vulnerabilities available for the JWT libraries in use.
```
Sjord: > It should be checked

I think passive voice like this should be avoided. Maybe it's better written as an instruction: "Check if there are any known vulnerabilities available for the JWT libraries in use."

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,98 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+
+#### Remediation
+
+The latest version available of the JWT libraries in use should be implemented, to avoid known vulnerabilities.
+
+Store the JWT on the mobile phone using a secure mechanism, like KeyChain on iOS or KeyStore on Android.
+
+If replay attacks are a risk for the app, `jti` (JWT ID) claim should be implemented.
+
+Ideally the content of JWT should be encrypted. An example implementation in Java can be found in the OWASP JWT Cheat Sheet<sup>[4]</sup>
```
Sjord: Maybe you can elaborate why you should encrypt the JWT.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,98 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+
+Several known vulnerabilities in JWT should be checked while executing a dynamic analysis:
+* Hashing algorithm `none`<sup>[6]</sup>:
+  * Modify the `alg` attribute in the token header and delete `HS256` and set it to `none` and use an empty signature (e.g. signature = ""). Use this token and replay it in a request. Some libraries treat tokens signed with the none algorithm as a valid token with a verified signature. This would allow an attacker to create their own "signed" tokens.
+* Usage of asymmetric algorithms:
+  * Within JWT asymmetric algorithms as RSA or ECDSA can be used for signing the token. Due to vulnerabilities in some JWT frameworks the public key can be misused as HMAC secret key in order to sign the token.
+* Token Storage on client side:
+  * When using a mobile app that uses JWT it should be verified where the token is stored locally on the device<sup>[5]</sup>.
+* Information Disclosure:
+  * Decode the Base-64 encoded JWT and check what kind of data is transmitted within it.
```
Sjord: Another possibility to compromise JWTs is to crack the HMAC's secret key.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,98 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+
+#### References
+
+##### OWASP Mobile Top 10 2016
+
+* M4 - Insecure Authentication - https://www.owasp.org/index.php/Mobile_Top_10_2016-M4-Insecure_Authentication
+
+##### OWASP MASVS
+
+* 4.1: "If the app provides users with access to a remote service, an acceptable form of authentication such as username/password authentication is performed at the remote endpoint."
```
Sjord: Is this really the correct requirement? I think username/password authentication gives you a JWT that can be used instead of a session token.

sushi2k: This requirement is just checking if some sort of authentication mechanism is used. Username/Password is just an example. What actually is used for authentication (cookie, JWT etc.) depends then on the implementation. 

These are the requirements for authentication, I think 4.1 fits best. What do you think?
https://github.com/OWASP/owasp-masvs/blob/master/Document/0x09-V4-Authentication_and_Session_Management_Requirements.md

sushi2k:
Thanks for your input @Sjord. Just updated it.

Document/0x07a-Testing-Authentication-and-Session-Management.md
Sjord: I don't know. It is sort of an alternative for 4.2. However, if you use a JWT instead of a session token you are violating 4.2, "the remote endpoint uses randomly generated access tokens".

Sjord: I opened an issue there: https://github.com/OWASP/owasp-masvs/issues/93

sushi2k: You are right the MASVS, this is not really covering stateless authentication. Thanks for opening the issue.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,106 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+```
+
+For mobile apps it's more and more used to authenticate both the message sender and receiver by using JWT. JWT implementations are available for all major programming languages, like PHP<sup>[2]</sup> or Java Spring<sup>[3]</sup>.
+
+#### Static Analysis
+
+Identify the JWT library that is used on server and client side. Check if there are any known vulnerabilities available for the JWT libraries in use.
+
+The following best practices should be verified in the JWT libraries<sup>[7]</sup>:
+* Verify the signature on server-side at all times for all incoming requests containing a token.
```
commjoen: signature or HMAC, and jot down that the keys should reside at the server.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,106 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+* Verify the data embedded into the JWT and check if sensitive information is available. If so, it should be considered to encrypt the JWT content.
```
commjoen: I rather ALWAYS encrypt the full content of the token OOORRR use Opaque tokens that refer to the actual token cached on a gateway.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,106 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+* Verify if replay attacks are addressed by using `jti` (JWT ID) claim, which provides a unique identifier for JWT.
+* Verify where the secret signing key is located and stored. It should only be available for the issuer and consumer.
```
commjoen: it's either the private signing key or the secret HMACing key. I would say that it is only available to the issuer and the verifier. Because saying consumer might mean that others believe they can use it too.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,106 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+
+#### Dynamic Analysis
+
+Several known vulnerabilities in JWT should be checked while executing a dynamic analysis:
+* Hashing algorithm `none`<sup>[6]</sup>:
+  * Modify the `alg` attribute in the token header and delete `HS256` and set it to `none` and use an empty signature (e.g. signature = ""). Use this token and replay it in a request. Some libraries treat tokens signed with the none algorithm as a valid token with a verified signature. This would allow an attacker to create their own "signed" tokens.
+* Usage of asymmetric algorithms:
+  * Within JWT asymmetric algorithms as RSA or ECDSA can be used for signing the token. Due to vulnerabilities in some JWT frameworks the public key can be misused as HMAC secret key in order to sign the token.
```
commjoen: These are two different items right? let's split them and formulate it a bit differently.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,106 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+* Token Storage on client side:
+  * When using a mobile app that uses JWT it should be verified where the token is stored locally on the device<sup>[5]</sup>.
+* Cracking the signing key:
+  * Creating a signature of the token is done through a secret key on server side. Once a JWT is obtained there are several tools available that can try to brute force the secret key offline<sup>[8]</sup>. See the tools section for details.
```
commjoen: secret key --> private key.

Document/0x07a-Testing-Authentication-and-Session-Management.md
commjoen: I think this issue should not be related to this item, but rather to 4.2 and stuff..

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -66,6 +66,106 @@ If any of these two conditions raise an issue, reject the request and do not all
...
+* [5] Sample of JWT Token - https://jwt.io/#debugger
+* [6] Critical Vulnerabilities in JSON Web Token - https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/
+* [7] JWT the right way - https://stormpath.com/blog/jwt-the-right-way
+* [8] Attacking JWT Authentication - https://www.sjoerdlangkemper.nl/2016/09/28/attacking-jwt-authentication/
+
+##### Tools
+* jwtbrute - https://github.com/jmaxxz/jwtbrute
+* crackjwt - https://github.com/Sjord/jwtcrack/blob/master/crackjwt.py
+* John the ripper - https://github.com/magnumripper/JohnTheRipper
+
```
commjoen: There are a couple items that need to be taken care off:
- signing out with a JWT (if logout has been allowed) should make the token invalid (e.g. blacklisted), which can be verified
- copying a token to another device should / or should not make an attacker able to continue authenticated
- nice-to-have: maybe check if we can relate it to other specs that are using jwt tokens? because this is mostly about how the token is setup, but not about how a token can be used in a context of oauth and stuff like that :).

sushi2k: > (if logout has been allowed) should make the token invalid 

This will be covered in the test case "### Testing the Logout Functionality". Will make a reference to it.

> copying a token to another device should / or should not make an attacker able to continue authenticated

Will make a reference to it.

> nice-to-have: maybe check if we can relate it to other specs that are using jwt tokens? because this is mostly about how the token is setup, but not about how a token can be used in a context of oauth and stuff like that :).

@cpholguera is working on a general description at the beginning of the chapter.

sushi2k:
Ok. Just pushed all the updates. Thanks for your input @Sjord and @commjoen. Let me know if something should still be fixed.

PR #537, opened 2017-06-14
Updated "Testing the Password Policy" subchapter

c0rdis:


Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -305,7 +305,13 @@ Regular Expressions are often used to validate passwords. The password verificat
 
 Passwords can be set when registering accounts, changing the password or when resetting the password in a forgot password process. All of the available functions in the application that are able to change or set a password need to be identified in the source code. They should all be using the same password verification check, that is aligned with the password policy.
 
-If a frameworks is used that offers the possibility to create and enforce a password policy for all users of the application, the configuration should be checked.
+Here are different examples on how a validation can be implemented server-side:
+
+* Spring (Java) -  https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/validation/Validator.html
+* Ruby on Rails -  http://guides.rubyonrails.org/active_record_validations.html
+* PHP - http://php.net/manual/en/filter.filters.validate.php
```
Sjord: In PHP at least the normal validation functions are rarely used to enforce password requirements. I think these resources are too focused on general input validation and not enough on password requirement validation.

[Zxcvbn](https://github.com/dropbox/zxcvbn) is a checker specifically for passwords. Other than that, some [custom code](https://stackoverflow.com/questions/10752862/password-strength-check-in-php) is pretty common in PHP.

Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -316,12 +322,12 @@ All available functions that allow a user to set a password need to be verified,
 - Forgot Password function that allows a user to set a new password or
 - Change Password function that allows a logged in user to set a new password.
 
-An interception proxy should be used, to bypass client passwords checks within the app in order to be able verify the password policy implemented on server side.
+An interception proxy should be used, to bypass client passwords checks within the app in order to be able verify the password policy implemented on server side. More information about testing methods can be found in the OWASP Testing Guide (OTG-AUTHN-007)<sup>[1]</sup>
```
Sjord: Is it really necessary to do password complexity checks on the server? I know client site checks can be bypassed, but this only allows a user to set his own password to something that doesn't match the password policy.

Document/0x07a-Testing-Authentication-and-Session-Management.md
c0rdis: Thanks, good point. I've added the section with these three to make the doc more consistent through the subchapters (and to show how things can be done "natively") but I agree that Zxcvbn or similar are more appropriate here.

Document/0x07a-Testing-Authentication-and-Session-Management.md
commjoen: I would recommend to make it necessary indeed 👍 , because you don't want to end up with "welcome12" as a password

sushi2k: Client side controls are always nice to have, but there needs to be a server side control in place to ensure consistency.

sushi2k:
Thanks for the update and the review. I will add the Zxcvbn

PR #547, opened 2017-06-15
Added OTP bruteforce + minor changes

c0rdis:
OTP bruteforce according to https://sakurity.com/otp. 
CWE links are now consistent.

### DIFF
--- a/Document/0x07a-Testing-Authentication-and-Session-Management.md
+++ b/Document/0x07a-Testing-Authentication-and-Session-Management.md
@@ -114,7 +114,7 @@ If any of these two conditions raise an issue, reject the request and do not all
 
 ##### CWE
 
-- CWE-287: Improper Authentication - https://cwe.mitre.org/data/definitions/287.html
+- CWE-287: Improper Authentication
 
 ##### Info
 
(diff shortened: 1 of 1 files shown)


Document/0x07a-Testing-Authentication-and-Session-Management.md
```diff
@@ -587,12 +587,16 @@ First, all privileged endpoints a user can only access with step-up authenticati
 
 The recorded requests should also be replayed without providing any authentication information, in order to check for a complete bypass of authentication mechanisms.
 
+Another attack is related to the case "Testing Excessive Login Attempts" - given that many OTPs are just numeric values, if the accounts are not locked after N unsuccessful attempts on this stage, an attacker can bypass second factor by simply bruterorcing the values within the range at the lifespan of the OTP. For 6-digit values and 30-second time step it would take less than 72 hours to find a match.
```
sushi2k: Good point, but how do you calculate 72 hours? If the OTP is 6 digits it's 1.000.000 combinations and if it's renewed every 30 seconds you have maybe 300 guesses per OTP if you are able to send 10 requests per second and then you start all over again with the next OTP. 
If the OTP is not  renewed and you are able to brute force the same OTP value, then you might be able to brute force it in a day or two depending on the amount of requests you can sent.

c0rdis: After our discussion, just changed to 90% probability but no calculation steps not to make it too detailed :P

PR #568, opened 2017-06-16
0x07b/cert pinning general

bolot:
* Move cert pinning overview and dynamic testing to the general Networking chapter.
* Expanded the overview, added gotchas.
* Work in progress

### DIFF
--- a/Document/0x07b-Testing-Network-Communication.md
+++ b/Document/0x07b-Testing-Network-Communication.md
@@ -233,7 +233,7 @@ Ensure that critical operations require at least one additional channel to confi
 
 #### Overview
 
-Certificate pinning allows to hard-code the certificate or parts of it into the app that is known to be used by the server. This technique is used to reduce the threat of a rogue CA and CA compromise. Pinning the server’s certificate takes the CA out of the game. Mobile apps that implement certificate pinning only can connect to a limited numbers of servers, as a small list of trusted CAs or server certificates are hard-coded in the application.
+Certificate pinning allows to hard-code the certificate or parts of it into the app that is known to be used by the server<sup>[1]</sup>. This technique is used to reduce the threat of a rogue CA and CA compromise. Pinning the server’s certificate takes the CA out of the game. Mobile apps that implement certificate pinning only can connect to a limited numbers of servers, as a small list of trusted CAs or server certificates are hard-coded in the application.
 
 Implementing certificate pinning is not a trivial task. Most applications will not benefit from using this technique and may introduce new vulnerabilities if they implement it incorrectly or without regard to operational challenges<sup>1</sup>:
 
@@ -242,7 +242,7 @@ Implementing certificate pinning is not a trivial task. Most applications will n
 - It may break SSL/TLS validation.
 - It can be circumvented by reverse engineering or on rooted/jailbroken devices.
 
-Even large organizations with mature processes and large devops staff are not immune to mistakes in implementing certificate pinning<sup>1</sup>. One high profile case involving a bank involved an expired intermediate certificate and caused outage to the customers.
+Even large organizations with mature processes and large devops staff are not immune to mistakes in implementing certificate pinning<sup>[2]</sup>. One high profile case involving a bank involved an expired intermediate certificate and caused outage to the customers.
 
 Once the challenges are understood and the proper processes are being implemented, important decisions have to be made:
 
@@ -308,5 +308,6 @@ Once you have extracted the certificate from the application (e.g. using Cycript
 
 ##### Info
 
-- [1] Pinning: Not as simple as it sounds, https://usmile.at/symposium/2017/program/kozyrakis
+- [1] OWASP - Certificate and Public Key Pinning: https://www.owasp.org/index.php/Certificate_and_Public_Key_Pinning
+- [2] Pinning: Not as simple as it sounds, https://usmile.at/symposium/2017/program/kozyrakis
 
--- a/Document/0x07b-Testing-Network-Communication.md
+++ b/Document/0x07b-Testing-Network-Communication.md
@@ -227,3 +227,34 @@ Ensure that critical operations require at least one additional channel to confi
 * [1] The Mobile Application Security Verification Standard - https://github.com/OWASP/owasp-masvs/blob/master/Document/0x03-Using_the_MASVS.md
 * [2] Infobip 2FA library - https://2-fa.github.io/libraries/android-library.html
 * [3] Google Authenticator for Android - https://github.com/google/google-authenticator-android
+
+
+### Testing SSL/TLS Pinning
+
+#### Overview
+
+Certificate pinning allows to hard-code the certificate or parts of it into the app that is known to be used by the server. This technique is used to reduce the threat of a rogue CA and CA compromise. Pinning the server’s certificate takes the CA out of the game. Mobile apps that implement certificate pinning only can connect to a limited numbers of servers, as a small list of trusted CAs or server certificates are hard-coded in the application.
+
+
+#### Static Analysis
+
+#### Dynamic Analysis
+
+
+#### Remediation
+
+
+#### References
+
+
+
+##### OWASP Mobile Top 10 2016
+* M3 - Insecure Communication - https://www.owasp.org/index.php/Mobile_Top_10_2016-M3-Insecure_Communication
+
+##### OWASP MASVS
+* V5.4 "The app either uses its own certificate store, or pins the endpoint certificate or public key, and subsequently does not establish connections with endpoints that offer a different certificate or key, even if signed by a trusted CA."
+
+##### CWE
+
+##### Info
+
--- a/Document/0x07b-Testing-Network-Communication.md
+++ b/Document/0x07b-Testing-Network-Communication.md
@@ -262,6 +262,34 @@ Refer to the platform specific networking chapter for additional details.
 
 #### Dynamic Analysis
 
+Dynamic analysis can be performed by launching a MITM attack using your preferred interception proxy<sup>[1]</sup>. This will allow to monitor the traffic exchanged between client (mobile application) and the backend server. If the Proxy is unable to intercept the HTTP requests and responses, the SSL pinning is correctly implemented.
+
+##### Server certificate validation
+
+We start our analysis by testing the application's behaviour while establishing secure connection. Our test approach is to gradually relax security of SSL handshake negotiation and check which security mechanisms are enabled.
+
+1. Having burp set up as a proxy in wifi settings, make sure that there is no certificate added to trust store (Settings -> General -> Profiles) and that tools like SSL Kill Switch are deactivated. Launch your application and check if you can see the traffic in Burp. Any failures will be reported under 'Alerts' tabl. If you can see the traffic, it means that there is no certificate validation performed at all! This effectively means that an active attacker can silently do MiTM against your application. If however, you can't see any traffic and you have an information about SSL handshake failure, follow the next point.
+2. Now, install Burp certificate, as explained in [Basic Security Testing section](./0x06b-Basic-Security-Testing.md). If the handshake is successful and you can see the traffic in Burp, it means that certificate is validated against device's trust store, but the pinning is not performed. The risk is less significant than in previous scenario, as two main attack scenarios at this point are misbehaving CAs and phishing attacks, as discussed in [Basic Security Testing section](./0x06b-Basic-Security-Testing.md).
+3. If executing instructions from previous step doesn't lead to traffic being proxied through burp, it means that certificate is actually pinned and all security measures are in place. However, you still need to bypass the pinning in order to test the application. Please refer to [Basic Security Testing section](./0x06b-Basic-Security-Testing.md) for more information on this.
+
+##### Client certificate validation
+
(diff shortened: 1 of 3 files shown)


Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
 * [1] The Mobile Application Security Verification Standard - https://github.com/OWASP/owasp-masvs/blob/master/Document/0x03-Using_the_MASVS.md
 * [2] Infobip 2FA library - https://2-fa.github.io/libraries/android-library.html
 * [3] Google Authenticator for Android - https://github.com/google/google-authenticator-android
+
+
+### Testing SSL/TLS Pinning
+
+#### Overview
+
+Certificate pinning allows to hard-code the certificate or parts of it into the app that is known to be used by the server<sup>[1]</sup>. This technique is used to reduce the threat of a rogue CA and CA compromise. Pinning the server’s certificate takes the CA out of the game. Mobile apps that implement certificate pinning only can connect to a limited numbers of servers, as a small list of trusted CAs or server certificates are hard-coded in the application.
```
Sjord: > Certificate pinning allows to hard-code the certificate or parts of it into the app that is known to be used by the server

I would place "that is known to be used by the server" closer to "certificate". Currently it looks like the app is used by the server.

> This technique is used to reduce the threat of a rogue CA and CA compromise.

I think it should be "or". Also, maybe you can swap the two so it is clear that you don't mean a "rogue CA compromise". So it should be something like "of a CA compromise or rogue CA."

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+
+Implementing certificate pinning is not a trivial task. Most applications will not benefit from using this technique and may introduce new vulnerabilities if they implement it incorrectly or without regard to operational challenges<sup>1</sup>:
+
+- It does not protect against compromises of the pinned certificate itself.
+- It makes the operations more complex and thus requires a more mature process.
+- It may break SSL/TLS validation.
+- It can be circumvented by reverse engineering or on rooted/jailbroken devices.
+
+Even large organizations with mature processes and large devops staff are not immune to mistakes in implementing certificate pinning<sup>[2]</sup>. One high profile case involving a bank involved an expired intermediate certificate and caused outage to the customers.
```
Sjord: > involving a bank involved

I think the "involve" twice is a bit confusing. Also, you probably need a citation to an article mentioning this outage.

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+
+Once the challenges are understood and the proper processes are being implemented, important decisions have to be made:
+
+- Which level in the certificate chain should be pinned, CA, intermediate, or leaf? Pinning to the most specific certificate reduces flexibility, pinning to the CA increases vulnerability.
```
Sjord: > pinning to the CA increases vulnerability

Is this correct? It is possible that you are totally secure if you pin to the CA. It increases the risk in some cases. Maybe this can be simplified to "Pinning to the most specific certificate reduces flexibility, but increases security."

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+- What should be pinned, the certificate, the certificate bytes, public key or the thumb?
```
Sjord: I know what the public key is, but of the others I am not so sure. Do you mean fingerprint with "thumb"?

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+- Lifecycle concerns: since the certificates expire, may be revoked or compromised, processes that handle these situations should be implemented.
+- Platform-specific implementation details are discussed in the respective networking chapters.
```
Sjord: I don't think this note should be in this list, as it is not one of the "important decisions" that have to be made.

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+If source code is available, the test should check for the following issues:
+
+* Apps performing their own X.509 certificate chain validation. They should instead rely on the existing proven implementations available on the platform SDKs or in trusted, updated third party libraries.
+* Apps implementing certificate pinning by using custom trust/CA stores. These are potentially error prone and may result in performing no certificate validation.
+
+Refer to the platform specific networking chapter for additional details.
+
+#### Dynamic Analysis
+
+Dynamic analysis can be performed by launching a MITM attack using your preferred interception proxy<sup>[1]</sup>. This will allow to monitor the traffic exchanged between client (mobile application) and the backend server. If the Proxy is unable to intercept the HTTP requests and responses, the SSL pinning is correctly implemented.
```
Sjord: > If the Proxy is unable to intercept the HTTP requests and responses, the SSL pinning is correctly implemented.

Is this really true? Couldn't there be something wrong with the pinning anyway when it is not possible to intercept the requests?

Also, I don't think "Proxy" should be capitalized here.

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+2. Install the Burp certificate, as explained in the Basic Security Testing section. If the handshake is successful and you can see the traffic in Burp, it means that certificate is validated against device's trust store, but the pinning is not performed. The risk is less significant than in the previous scenario, as two main attack scenarios at this point are misbehaving CAs and phishing attacks, as discussed in Basic Security Testing section.
+3. If executing instructions from previous step doesn't lead to traffic being proxied through Burp, it means that certificate is actually pinned and all security measures are in place. However, you still need to bypass the pinning in order to test the application. Please refer to Basic Security Testing section for more information on this.
+
+##### Client certificate validation
+
+Some applications use a two-way SSL handshake, meaning that the application verifies server's certificate and the server verifies the client's certificate. You can notice this if there is an error in Burp 'Alerts' tab indicating that client failed to negotiate connection.
```
Sjord: > two-way SSL handshake

I am not sure whether this is the correct term for this. I think this should be "mutual authentication" or "two-way authentication".

> verifies server's

This should be "verifies the server's"

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+
+There are a couple of things worth noting:
+
+1. Client certificate contains private key that will be used in key exchange.
+2. Usually certificate would also need a password to use (decrypt) it.
+3. Certificate itself can be stored in the binary itself, data directory or the keychain.
```
Sjord: There are some "a"s and "the"s missing here.

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+
+Most common but improper and insecure way of doing two-way handshake is to store client certificate within the application bundle and hardcode the password. This obviously does not bring much security, because all clients will share the same certificate.
+
+Second way of storing the certificate (and possibly password) is to use the keychain. Upon first login, the application should download personal certificate and store it securely in the keychain.
+
+Sometimes application have one certificate that is hardcoded and used for first login and then personal certificate is downloaded. In this case, check if it's possible to still use the 'generic' certificate to connect to the server.
```
Sjord: > Sometimes application have one certificate

Sometimes _the_ application _has_ one certificate

sushi2k:
Hey @bolot. The content looks great and Sjord already made some good suggestions. 
I just looked at the structure again and I think we should make two separate test cases for iOS and Android. The intention of moving test cases to the generic section was because the logic/code that need to be checked is mostly on server side and therefore the implementation in iOS or Android is not in scope for testing. Which is the case for the generic authentication, network, cryptography chapters that we have at the moment. 

For SSL pinning this is different, as the app need to implement the whole pinning logic. There are different libraries that can be used on Android (which is already listed in the Android test case) and different implementation in iOS (objctive-c and Swift). Therefore I would suggest to create two test cases and move the content from the generic test  case back to the iOS/Android chapter. Otherwise it's too fragmented and hard to read due to the references. There will be overlaps in the dynamic  and also in the recommendation section, but the reader will definitely have it easier to execute the steps in the test case. What do you think?

sushi2k:
For the dynamic part we could also refer to SSLKillSwitch for iOS and Android-SSL-TrustKiller on Android. Therefore the static and dynamic part would be unique to the platform besides the Mitm attack.

Document/0x07b-Testing-Network-Communication.md
commjoen: both terms are used, maybe add them both?

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+- Which level in the certificate chain should be pinned, CA, intermediate, or leaf? Pinning to the most specific certificate reduces flexibility, pinning to the CA increases vulnerability.
+- What should be pinned, the certificate, the certificate bytes, public key or the thumb?
+- Lifecycle concerns: since the certificates expire, may be revoked or compromised, processes that handle these situations should be implemented.
+- Platform-specific implementation details are discussed in the respective networking chapters.
+
+#### Static Analysis
+
+If source code is available, the test should check for the following issues:
+
+* Apps performing their own X.509 certificate chain validation. They should instead rely on the existing proven implementations available on the platform SDKs or in trusted, updated third party libraries.
```
commjoen: Perhaps change end of first sentence into: "chain validation and cleanup"

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+2. Install the Burp certificate, as explained in the Basic Security Testing section. If the handshake is successful and you can see the traffic in Burp, it means that certificate is validated against device's trust store, but the pinning is not performed. The risk is less significant than in the previous scenario, as two main attack scenarios at this point are misbehaving CAs and phishing attacks, as discussed in Basic Security Testing section.
+3. If executing instructions from previous step doesn't lead to traffic being proxied through Burp, it means that certificate is actually pinned and all security measures are in place. However, you still need to bypass the pinning in order to test the application. Please refer to Basic Security Testing section for more information on this.
```
commjoen: it could be as well that there is a difference in behavior: e.g. the pinning is broken, but traffic is being send through. One might have just informed the server that the pinning is broken by sending additional headers or calling some endpoint it would normally never call.

Document/0x07b-Testing-Network-Communication.md
```diff
@@ -227,3 +227,87 @@ Ensure that critical operations require at least one additional channel to confi
...
+
+There are a couple of things worth noting:
+
+1. Client certificate contains private key that will be used in key exchange.
+2. Usually certificate would also need a password to use (decrypt) it.
+3. Certificate itself can be stored in the binary itself, data directory or the keychain.
+
+Most common but improper and insecure way of doing two-way handshake is to store client certificate within the application bundle and hardcode the password. This obviously does not bring much security, because all clients will share the same certificate.
+
+Second way of storing the certificate (and possibly password) is to use the keychain. Upon first login, the application should download personal certificate and store it securely in the keychain.
```
commjoen: bear in mind: a keychain in android does something else than the keychain in ios :D.

bolot:
@sushi2k I'll re-work the PR based on your recommendations, taking into account the comments from @Sjord and @commjoen. Expected update: mid-next week

Document/0x07b-Testing-Network-Communication.md
pmilosev: If the proxy certificate (self signed) is not added to test device CA store, then the connection will fail even without pinning. Probably here we should have a reference to how to setup proxy, etc.
Also the observed behavior should be that no SSL / TLS connection can be established, in which case there is no HTTP request / response to intercept.
If you see a HTTP request, even if you can not intercept if or decrypt it, then the pinning is not done properly.

Document/0x07b-Testing-Network-Communication.md
pmilosev: I've observed the term 'two way SSL' quite more often than any other term for it.
By the way, 'mutual authentication' and'two way authentication' even seem wrong, as there are different ways to authenticate (e.g. basic authentication'). Two way SSL, or client certificate authentication refers to one specific kind of client authentication.

Sjord: > a two-way SSL handshake

Maybe you could remove "handshake" from this to make it more clear?

b-mueller:
@bolot has this been updated?

bolot:
Let me close this PR and open a separate one.

PR #571, opened 2017-06-16
OAuth2 lack of verification attack scheme

c0rdis:
"3rd-party apps simply (and incorrectly) login the user based on the received user-id WITHOUT verifying whether the received user-id is indeed bound to the issued OAuth access token"


sushi2k:
Will be waiting for another commit to get an explanation for the picture. Did you upload the picture to Document/Images/Chapters/0x07a? I cannot see the target path for the picture in the PR.

c0rdis:
Explanation: "On the step 3 during OAuth2 message exchange, the attacker substitutes her own user id (user id in the IdP or email address) with the victim’s one. If the backend server directly uses the user's identity proof returned from the client-side app to identify the user WITHOUT further validation whether the received user id is bound to the issued OAuth2 access token, the attacker can successfully sign into the app as the victim and have full access to the victim's sensitive information on our server".

sushi2k:
I merge this now. Please use the explanation then in your next PR. Thanks

Issue #467, opened 2017-05-20
[EDITING] Chapter "Testing Data Storage on Android"


b-mueller:
See here:

https://github.com/OWASP/owasp-mstg/blob/master/Document/0x05d-Testing-Data-Storage.md

pmilosev:
I'm starting the review.

b-mueller:
Thanks for the review so far. I added some more changes.

Still, I think that large parts of this chapter need to be rewritten due to grammar errors and style issues. IMHO we need to get someone with good English & writing skills to completely overhaul it.

sushi2k:
@pmilosev Any update on this or any timeline when you might be able to review? Thanks

pmilosev:
I should be able to finish the review next week.

sushi2k:
Great. There were a lot of updates in the guide during the OWASP Summit, so also do a pull request first ;-)

pmilosev:
I finished the review.
I agree with @b-mueller that this chapter needs some better wording.
Other than that the content looks OK.
I just added a small hint on potential issues when stripping logs with ProGuard - see the PR above.

pmilosev:
p.s. I personally (as a user of a mobile device) didn't like the idea of the mobile app gaining admin rights to check if proper security is ensured by the user (e.g. if strong device password is set, etc.).
Personal freedom (privacy, freedom of choice, etc.) is already secondary to security for many application providers (and some are even using security only as an excuse). I'm afraid that having OWASP suggest that it's OK to get some freedom out of the hands of the user may be badly misunderstood / misused by many.
I'm well aware of why this requirement is there, but I would rather be quite careful with the wording :)

pmilosev:
@b-mueller, is there something I should do about this ticket, as I finished the review ?

PR #583, opened 2017-06-21
Added reference to an issue regarding log stripping with ProGuard.

pmilosev:
Also offered an alternative approach based on commenting out log statements with Gradle.
#467

### DIFF
--- a/Document/0x05d-Testing-Data-Storage.md
+++ b/Document/0x05d-Testing-Data-Storage.md
@@ -414,8 +414,26 @@ public static int e(...);
 public static int wtf(...);
 }
 ```
-Please note that the above example only ensures the logging calls will be stripped. Any code for dynamic log statement generation, might be left in the bytecode, making the log statements accessible with a debugger <sup>[3]</sup>. According to the ProGuard author, latest version of ProGuard implement some heuristics to improve on this issue, however there is no guarantee, that complex string builders will be removed.
-To avoid the issue, one may choose to remove logs on source level, instead of compiled bytecode. Bellow is a simple Gradle task which comments out all log statements including any inline string builder.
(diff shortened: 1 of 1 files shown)


Document/0x05d-Testing-Data-Storage.md
```diff
@@ -414,6 +414,23 @@ public static int e(...);
 public static int wtf(...);
 }
 ```
+Please note that the above example only ensures the logging calls will be stripped. Any code for dynamic log statement generation, might be left in the bytecode, making the log statements accessible with a debugger <sup>[3]</sup>. According to the ProGuard author, latest version of ProGuard implement some heuristics to improve on this issue, however there is no guarantee, that complex string builders will be removed.
```
Sjord: > Any code for dynamic log statement generation, might be left in the bytecode, making the log statements accessible with a debugger.

I think the first comma should be removed.

> According to the ProGuard author, latest version of ProGuard implement some heuristics to improve on this issue, however there is no guarantee, that complex string builders will be removed.

This is a bit of a long sentence. I think it would be better if you split in two.

> only ensures the logging calls will be stripped. Any code for dynamic log statement generation, might be left in the bytecode.

After reading this it was not clear to me what the problem was. Maybe you can elaborate on it some more. Explain that it removes the call to the log function, but not the parameters to it, or something like that.

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -414,6 +414,23 @@ public static int e(...);
...
+To avoid the issue, one may choose to remove logs on source level, instead of compiled bytecode. Bellow is a simple Gradle task which comments out all log statements including any inline string builder.
```
Sjord: > Bellow

Below

b-mueller:
Thanks @pmilosev!

PR #585, opened 2017-06-22
Update 0x08-Testing-Tools.md

bugwrangler:


b-mueller:
Thanks for adding all the examples and screenshots. We can't merge it as-is because there's some grammar issues and typos, and some of the content would fit better into "0x06h - Testing Platfom Interaction:. But I'll try to work it in with some rearrangements and fixes.

PR #607, opened 2017-06-29
Restore the ECB mode test

bolot:
The ECB mode test was removed in commit 87a70f671782911a9124693c4b1dc0451d9ee2dc as part of the general cryptography section clean up. At the time the plan was to pare down the general section in favor of moving some content to the platform-specific chapters. However, as we were going through the clean up, we realized that the general crypto chapter should remain and performed some updates to it. Unfortunately, the ECB mode test was one of the casualties that wasn't restored.


Sjord:
Looks good to me. Maybe you can use the [new reference style](https://github.com/OWASP/owasp-mstg/issues/590) and remove the info section.

sushi2k:
Thanks Bolot

Issue #590, opened 2017-06-23
Inline References


b-mueller:
Convert all references to inline.

For web links:

\[TEXT](URL "NAME")

For example:

The [threat modeling guidelines defined by OWASP](https://www.owasp.org/index.php/Application_Threat_Modeling "OWASP Application Threat Modeling") are generally applicable to mobile apps.

For books and papers:

\[#NAME]

And include the full reference in the "references" section manually:

E.g.:

An obfuscated encryption algorithm can generate its key (or part of the key) using data collected from the environment [#riordan].

And under "references" at the end of the document:

- [#riordan] -  James Riordan, Bruce Schneier. Environmental Key Generation towards Clueless Agents. Mobile Agents and Security, Springer Verlag, 1998

TheDauntless:
What about references where there's no good text to refer too? For example:

> This reveals another surface of attacks aimed at data on the way. It's possible for an attacker to sniff or even modify (MiTM attacks) an unencrypted information if he controls any part of network infrastructure (e.g. an WiFi Access Point)<sup>[1]</sup>.

sushi2k:
In this case the reference can be removed from 0x07b-Testing-Network-Communication). Otherwise we need to find a way to get a text :-)

sushi2k:
Done:

[x] Android Data Storage (https://github.com/OWASP/owasp-mstg/blob/master/Document/0x05d-Testing-Data-Storage.md)

b-mueller:
The following chapters still have old style references:

https://github.com/OWASP/owasp-mstg/blob/master/Document/0x05i-Testing-Code-Quality-and-Build-Settings.md
https://github.com/OWASP/owasp-mstg/blob/master/Document/0x06d-Testing-Data-Storage.md
https://github.com/OWASP/owasp-mstg/blob/master/Document/0x06g-Testing-Network-Communication.md
https://github.com/OWASP/owasp-mstg/blob/master/Document/0x06j-Testing-Resiliency-Against-Reverse-Engineering.md
https://github.com/OWASP/owasp-mstg/blob/master/Document/0x07c-Testing-Cryptography.md
https://github.com/OWASP/owasp-mstg/blob/master/Document/0x07d-Assessing-Anti-Reverse-Engineering-Schemes.md

Any volunteers? :)

sushi2k:
Let's finish it up next week. I try to do one per day :-)

sushi2k:
Done: 
[...]

PR #610, opened 2017-06-29
Modify pages for inline references

TheDauntless:
See issue #590


sushi2k:
Great. Thanks for the PR

PR #629, opened 2017-07-03
 Remove the caps-rule from the styleguide 

Sjord:
Don't use CAPS for emphasis. I found only one place where we use this. Instead
we should just use _markdown_ **emphasis**.


sushi2k:
Makes sense. Thanks

PR #646, opened 2017-07-06
replace is with are

feedthedogs:


Sjord:
Thanks, @feedthedogs.

PR #654, opened 2017-07-10
replace 'fast' by 'fact' in 'the fact is' sentence

LucasParsy:
On the second line of the diff, there is a typo, in "However, the fast is that",
I think you mean "the fact is".

It is just a small fix, but I'd be glad to become an  °˖✧OWASP Contributor✧˖°!


b-mueller:
Thanks @LucasParsy :)

PR #661, opened 2017-07-11
Update all references to MASVS

Sjord:
Make them consistent according to the testcase template.


sushi2k:
Love your script. Now the MASVS references are consistent. Thanks for the PR

PR #666, opened 2017-07-14
you can detect some cryptographic issues without having access to the…

andreashappe:
… source code

### DIFF
--- a/Document/0x04g-Testing-Cryptography.md
+++ b/Document/0x04g-Testing-Cryptography.md
@@ -85,7 +85,7 @@ Other uses of cryptography require careful adherence to best practices:
 
 #### Dynamic Analysis
 
-If passwords are hashed with an insecure algorithm with known collisions an attacker can utilize this to [gain knowledge about the used hash algorithm](https://www.netsparker.com/blog/web-security/collision-based-hashing-algorithm-disclosure/). For example, the pair `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa**200**a8284bf36e8e4b55b35f427593d849676da0d1**555**d8360fb5f07fea2` and `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa**202**a8284bf36e8e4b55b35f427593d849676da0d1**d55**d8360fb5f07fea2` both produce the same MD5 sum. If an attacker can set his password to the former and is then able to login using the latter as a password, usage of the insecure MD5 algorithm as hash algorithm has been confirmed.
+If passwords are hashed with an insecure algorithm with known collisions an attacker can utilize this to [gain knowledge about the used hash algorithm](https://www.netsparker.com/blog/web-security/collision-based-hashing-algorithm-disclosure/). For example, the pair <tt>4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa<b>200</b>a8284bf36e8e4b55b35f427593d849676da0d1<b>555</b>d8360fb5f07fea2</tt> and <tt>4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa<b>202</b>a8284bf36e8e4b55b35f427593d849676da0d1<b>d55</b>d8360fb5f07fea2</tt> both produce the same MD5 sum. If an attacker can set his password to the former and is then able to login using the latter as a password, usage of the insecure MD5 algorithm as hash algorithm has been confirmed.
 
 #### Remediation
 
--- a/Document/0x04g-Testing-Cryptography.md
+++ b/Document/0x04g-Testing-Cryptography.md
@@ -85,7 +85,7 @@ Other uses of cryptography require careful adherence to best practices:
 
 #### Dynamic Analysis
 
-If passwords are hashed with an insecure algorithm with known collisions an attacker can utilize this to [gain knowledge about the used hash algorithm](https://www.netsparker.com/blog/web-security/collision-based-hashing-algorithm-disclosure/). For example, the pair `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa200a8284bf36e8e4b55b35f427593d849676da0d1555d8360fb5f07fea2` and `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa202a8284bf36e8e4b55b35f427593d849676da0d1d55d8360fb5f07fea2` both produce the same MD5 sum. If an attacker can set his password to the former and is then able to login using the latter as a password, usage of the insecure MD5 algorithm as hash algorithm has been confirmed.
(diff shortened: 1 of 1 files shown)


Document/0x04g-Testing-Cryptography.md
```diff
@@ -83,6 +83,10 @@ Other uses of cryptography require careful adherence to best practices:
 - RSA with 768-bit and weaker keys can be broken. Older PKCS#1 padding leaks information.
 - Rely on secure hardware, if available, for storing encryption keys, performing cryptographic operations, etc.
 
+#### Dynamic Analysis
+
+If passwords are hashed with an insecure algorithm with known collisions an attacker can utilize this to [gain knowledge about the used hash algorithm](https://www.netsparker.com/blog/web-security/collision-based-hashing-algorithm-disclosure/). For example, the pair `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa200a8284bf36e8e4b55b35f427593d849676da0d1555d8360fb5f07fea2` and `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa202a8284bf36e8e4b55b35f427593d849676da0d1d55d8360fb5f07fea2` both produce the same MD5 sum. If an attacker can set his password to the former and is then able to login using the latter as a password, usage of the insecure MD5 algorithm as hash algorithm has been confirmed.
```
sushi2k: That's a nice trick. Was not aware of this. Can you please highlight the difference and make it bold? It took me a while to see where the difference is.

Document/0x04g-Testing-Cryptography.md
```diff
@@ -130,6 +134,14 @@ Through source code analysis the following non-exhausting configuration options
 - Fit-for-purpose block encryption modes
 - Key management being done properly
 
+#### Dynamic Analysis
+
+Various weaknesses can be detected within out access to the source code:
```
sushi2k: "within out"  -> "without"

Document/0x04g-Testing-Cryptography.md
```diff
@@ -130,6 +134,14 @@ Through source code analysis the following non-exhausting configuration options
...
+
+- if access to encrypted files or to generated hashes is possible (e.g., through stored files) the same file can be encrypted by two users. If this results in the same encrypted file, a static encryption key has been used.
+- if two different users are able to generate the same hash for the same file, then no (or an indequate) salt has been used.
+- IVs are usually stored in plain-text at the beginning of a file; compare differnt encrypted filesand check if their encrypted representation does not contain an identical salt.
```
sushi2k: "filesand" -> "files and"

Document/0x04g-Testing-Cryptography.md
```diff
@@ -130,6 +134,14 @@ Through source code analysis the following non-exhausting configuration options
 - Fit-for-purpose block encryption modes
 - Key management being done properly
 
+#### Dynamic Analysis
+
+Various weaknesses can be detected within out access to the source code:
+
+- if access to encrypted files or to generated hashes is possible (e.g., through stored files) the same file can be encrypted by two users. If this results in the same encrypted file, a static encryption key has been used.
```
Sjord: I think this sentence could be made a bit clearer. I was confused by the first sentence:

> if access to encrypted files or to generated hashes is possible (e.g., through stored files) the same file can be encrypted by two users.

You have here an "if a, then b", but b is not dependant on a. The same file can be encrypted by two users whether access to the encrypted files is possible or not. Furthermore, this does not indicate a weakness, which only comes in the next sentence. I would change it to something like this:

> Let two users encrypt a file. If this results in the same encrypted file, the app uses a static encryption key. Access to the encrypted file is necessary to confirm this weakness.

Document/0x04g-Testing-Cryptography.md
```diff
@@ -130,6 +134,14 @@ Through source code analysis the following non-exhausting configuration options
...
+- if two different users are able to generate the same hash for the same file, then no (or an indequate) salt has been used.
```
Sjord: This looks very similar to the first point. Does this deserve its own bullet point?

Document/0x04g-Testing-Cryptography.md
Sjord: You mix up IV and salt. While it is true they serve a similar purpose, these terms are not interchangable.

Also, differnt -> different

Document/0x04g-Testing-Cryptography.md
andreashappe: Good question. I would keep them separated as they attack different cryptographic directives but can change them if you want to.

Document/0x04g-Testing-Cryptography.md
andreashappe: Yes, true: I changed the final salt into IVs. What do you think about adding a reference to http://cwe.mitre.org/data/definitions/329.html or http://www.cryptofails.com/post/70059609995/crypto-noobs-1-initialization-vectors ?

Document/0x04g-Testing-Cryptography.md
```diff
@@ -83,6 +83,10 @@ Other uses of cryptography require careful adherence to best practices:
 - RSA with 768-bit and weaker keys can be broken. Older PKCS#1 padding leaks information.
 - Rely on secure hardware, if available, for storing encryption keys, performing cryptographic operations, etc.
 
+#### Dynamic Analysis
+
+If passwords are hashed with an insecure algorithm with known collisions an attacker can utilize this to [gain knowledge about the used hash algorithm](https://www.netsparker.com/blog/web-security/collision-based-hashing-algorithm-disclosure/). For example, the pair `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa**200**a8284bf36e8e4b55b35f427593d849676da0d1**555**d8360fb5f07fea2` and `4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa**202**a8284bf36e8e4b55b35f427593d849676da0d1**d55**d8360fb5f07fea2` both produce the same MD5 sum. If an attacker can set his password to the former and is then able to login using the latter as a password, usage of the insecure MD5 algorithm as hash algorithm has been confirmed.
```
Sjord: > `...bfa**200**a82...0d1**555**d836...`

The markdown bold `**` formatting doesn't working within a code block. I am not sure how to highlight a piece of code.

Document/0x04g-Testing-Cryptography.md
```diff
@@ -130,6 +134,14 @@ Through source code analysis the following non-exhausting configuration options
 - Fit-for-purpose block encryption modes
 - Key management being done properly
 
+#### Dynamic Analysis
+
+Various weaknesses can be detected without access to the source code:
+
+- Let two users encrypt a file. If this results in the same encrypted file, the app uses a static encryption key. Access to the encrypted file is necessary to confirm this weakness.
+- if two different users are able to generate the same hash for the same file, then no (or an indequate) salt has been used.
+- IVs are usually stored in plain-text at the beginning of a file; compare different encrypted files and check if their encrypted representation does not contain an identical IVs.
```
Sjord: > an identical IVs

This should be either "identical IVs", or "an identical IV".

Document/0x04g-Testing-Cryptography.md
andreashappe: I hoped that it might work (seems dependent upon the markdown engine). I could try to use HTML bold tags (<b>).. do you know if those might work?

Sjord: You can combine the `<tt>` and `<b>` HTML tags. `<tt>test<b>test</b>test</tt>` results in <tt>test<b>test</b>test</tt>.

Document/0x04g-Testing-Cryptography.md
Sjord: I think it would be a good idea to reference the cryptofails post to provide more insight into the purpose of IV's. You can add the CWE to the CWE section below.

Sjord:
👍 Can I merge this, @andreashappe ?

PR #669, opened 2017-07-15
TLS error ignoring in the context of WebView

righettod:
Hello,

I have added information about TLS error ignoring in the context of WebView, precisely in case of bad usage of the WebViewClient.

Thanks in advance for your feedback


sushi2k:
Thanks, good point the webview was not covered yet.

PR #671, opened 2017-07-16
SSL Pinning for WebView and NSC

righettod:
Hello,

I have added information about SSL Pinning in the case of usage of:
* WebView
* Network Security Config feature of Android v7+

Thanks in advance

### DIFF
--- a/Document/0x05g-Testing-Network-Communication.md
+++ b/Document/0x05g-Testing-Network-Communication.md
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
         .build();
 ```
 
+Applications that use WebView component may utilize event handler of the WebViewClient in order to perform some kind of "certificate pinning" on each request before that the target resource will be loaded. Verification code seems like the following:
+
+```java
+WebView myWebView = (WebView) findViewById(R.id.webview);
+myWebView.setWebViewClient(new WebViewClient(){
+    private String expectedIssuerDN = "CN=Let's Encrypt Authority X3,O=Let's Encrypt,C=US;";
+
+    @Override
+    public void onLoadResource(WebView view, String url)  {
+        //From Android API documentation about "WebView.getCertificate()":
+        //Gets the SSL certificate for the main top-level page or null if there is no certificate (the site is not secure).
+        //
+        //Available information on SslCertificate class are "Issuer DN", "Subject DN" and validity date helpers
+        SslCertificate serverCert = view.getCertificate();
+        if(serverCert != null){
+            //Apply check on Issuer DN against expected one
+            SslCertificate.DName issuerDN = serverCert.getIssuedBy();
+            if(!this.expectedIssuerDN.equals(issuerDN.toString())){
+                //Throw exception to cancel resource loading...
+            }
+        }
+    }
+});
+```
+
+Applications can decide to use the [Network Security Configuration](https://developer.android.com/training/articles/security-config.html) feature provided by Android from version 7.0.
+From Android documentation:
+"The Network Security Configuration feature lets apps customize their network security settings in a safe, declarative configuration file without modifying app code".
+
+Network Security Configuration (NSC) feature can also be used to perform [declarative certificate pinning](https://developer.android.com/training/articles/security-config.html#CertificatePinning) on specific domains. If an application uses the NSC feature then there 2 points to check in order to identify the defined configuration:
+
+1. Specification of the NSC file reference in the Android application manifest using the "android:networkSecurityConfig" attribute on the application tag:
+```xml
+<?xml version="1.0" encoding="utf-8"?>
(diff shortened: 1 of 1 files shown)


Document/0x05g-Testing-Network-Communication.md
```diff
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
         .build();
 ```
 
+Applications that use WebView component may utilize event handler of the WebViewClient in order to perform some kind of "certificate pinning" on each request before that the target resource will be loaded. Verification code seems like the following:
```
sushi2k: Applications that use **a** WebView component may utilize **the** event handler of the WebViewClient in order to perform some kind of "certificate pinning" on each request before that the target resource will be loaded. 

-> Please add the things I've put in bold
-> Please remove "that" in "that the target"

Instead of "Verification code seems like the following:" -> "The following code shows an example for verifying the Issuer DN of the certificate sent by the server. "

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+            SslCertificate.DName issuerDN = serverCert.getIssuedBy();
+            if(!this.expectedIssuerDN.equals(issuerDN.toString())){
+                //Throw exception to cancel resource loading...
+            }
+        }
+    }
+});
+```
+
+Applications can decide to use the [Network Security Configuration](https://developer.android.com/training/articles/security-config.html) feature provided by Android from version 7.0.
```
sushi2k: Please check the format for external references and remove the space between "] ("
https://github.com/OWASP/owasp-mstg/blob/master/style_guide.md#3-external-references

"from version 7.0" -> "from version 7.0 onwards, to customize their network security settings in a safe, declarative configuration file without modifying app code"."

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+From Android documentation:
+"The Network Security Configuration feature lets apps customize their network security settings in a safe, declarative configuration file without modifying app code".
```
sushi2k: I would remove this, as we don't want to just copy documentation and put a part of it in the sentence before instead. There you also have the reference.

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+
+Network Security Configuration (NSC) feature can also be used to perform [declarative certificate pinning](https://developer.android.com/training/articles/security-config.html#CertificatePinning) on specific domains. If an application uses the NSC feature then there 2 points to check in order to identify the defined configuration:
```
sushi2k: -> "The Network..." 

Please update the reference to the sytle guide shared above. 

-> "the there are two"

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+            <!-- Hash of the public key (SubjectPublicKeyInfo of the X.509 certificate) of the Intermediate CA of the OWASP website server certificate -->
+            <pin digest="SHA-256">YLh1dUR9y6Kja30RrAn7JKnbQG/uEtLMkBgFF2Fuihg=</pin>
+            <!-- Hash of the public key (SubjectPublicKeyInfo of the X.509 certificate) of the Root CA of the OWASP website server certificate -->
+            <pin digest="SHA-256">Vjs8r4z+80wjNcr1YKepWQboSIRi63WsWXhIMN+eWys=</pin>
+        </pin-set>
+    </domain-config>
+</network-security-config>
+```
+
+If a NSC configuration is in place then the following event can be visible in Log:
```
sushi2k: -> "in the log file:"

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -169,6 +169,74 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+```
+D/NetworkSecurityConfig: Using Network Security Config from resource network_security_config
+```
+
+If a certification pinning validation fail then the following event can be visible in Log:
```
sushi2k: -> "If a certificate pinning validation check is failing then the following event will be logged:"

sushi2k:
Thanks for sharing. That is good content. Please have a look at the suggestions.

righettod:
All feebacks has been taken in account in commit 5d57040 and i have adapted code snippet to avoid horizontal scrolling.

sushi2k:
Thanks Dominique. Will merge it now.

Issue #478, opened 2017-05-24
0x5d Test Case: Update Testing for Sensitive Data in Memory on Android


sushi2k:
https://github.com/OWASP/owasp-mstg/blob/master/Document/0x05d-Testing-Data-Storage.md#testing-for-sensitive-data-in-memory

sushi2k:
Hi @pmilosev. Any update for this test case?

Naushad007:
Testing for Sensitive Data in Local Storage...SQLite Databases

commjoen:
@Naushad007 is that not more on sensitive data in storage? Because on storage you would need to test for:
- SQLite
- SQLCipher 
- Realm
- Plists
- NSUserDefaults
...and gimme some more ;-)

Naushad007:
Thanks @commjoen I will follow up that streem

sushi2k:
Hi @Naushad007. This issue is about sensitive data exposed in the memory, not the file system. 

Do you want to work on this ticket? If so, please create a PR and I can assign the ticket also to you. I think @pmilosev is still busy at the moment and couldn't find time to update the test case.

pmilosev:
Sorry, few escalations at work and the plan goes out of window.
I'm on a vacation in the next few weeks, so I'll try to do at least some contribution, but let's not promise much.
I still have this project on my agenda though ;)

sushi2k:
Noted and understand. Just wanted to ping you :-)

pmilosev:
I've started reviewing and editing this chapter.

pmilosev:
done, check the linked pull request #675

PR #675, opened 2017-07-18
Edited and extended the "Testing for sensitive data in memory" chapter.

pmilosev:
Not yet ready to be merged!
I will continue reviewing and editing this chapter in the next days.
I opened the PR for early access to the changes (feel free to comment) and coordination if anyone else is looking in the issue.

#478

### DIFF
--- a/Document/0x05d-Testing-Data-Storage.md
+++ b/Document/0x05d-Testing-Data-Storage.md
@@ -1069,12 +1069,13 @@ setContentView(R.layout.activity_main);
 
 #### Overview
 
-Analyzing memory can help developers to identify root causes of different problems, like for example application crashes, etc. However, it can also be used to gain access to sensitive data. This section describes how to check for disclosure of data within the process' memory.
+Analyzing memory can help developers to identify root causes of several problems, such as application crashes. However, it can also be used to gain access to sensitive data. This section describes how to check for disclosure of data within the process' memory.
 
-First, you need to identify which sensitive information is stored in memory. Basically, if you have a sensitive asset it's very likely that at some point it is loaded in memory. The objective is to verify that these info has as short exposure as possible. 
+First, you need to identify which sensitive information is stored in memory. Basically, if you have a sensitive asset it's very likely that at some point it is loaded in memory. The objective is to verify that this info is exposed as briefly as possible.
 
-To be able to investigate the memory of an application a memory dump needs to be created first or the memory needs to be viewed with real-time updates. This is also already the problem, as the application only stores certain information in memory if certain functions are triggered within the application. Memory investigation can of course be executed randomly in every stage of the application, but it is much more beneficial to understand first what the mobile app is doing and what kind of functionalities it offers and also make a deep dive into the (decompiled) source code before making any memory analysis.
-Once sensitive functions are identified, like decryption of data, the investigation of a memory dump might be beneficial in order to identify sensitive data like a key or the decrypted information itself.
+To be able to investigate the memory of an application a memory dump needs to be created first. Alternatively it can be analyzed in real-time, e.g. over a debugger. No matter the approach, this is a very error prone process from a verification point of view, as what you will get in a certain dump is the data left by the functions that were executed. You might miss executing critical scenarios. Additionally, unless you know the footprint of the data you are looking for (either the exact value, or its format), it is quite easy not to identify it during analysis. For example, if the app performs encryption based on a randomly generated symmetric key, unless you get to know the value of the key by other means, it is very unlikely that you will be able to spot it in memory.
+
+Therefore you are better off starting with static analysis.
 
 #### Static Analysis
 
--- a/Document/0x05d-Testing-Data-Storage.md
+++ b/Document/0x05d-Testing-Data-Storage.md
@@ -1093,7 +1093,7 @@ Unfortunately, not many libraries and frameworks are designed to allow overwriti
 
 ```java
 SecretKey secretKey = new SecretKeySpec("key".getBytes(), "AES");
-secret.destroy();
+secretKey.destroy();
 ```
 
 Neither does overwriting the backing byte-array from `secretKey.getEncoded()` as the SecretKeySpec based key returns a copy of the backing byte-array. Take a look in the remediation section below on how to properly remove a `SecretKey` from memory.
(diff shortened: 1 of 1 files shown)


Document/0x05d-Testing-Data-Storage.md
```diff
@@ -1069,22 +1069,35 @@ setContentView(R.layout.activity_main);
 
 #### Overview
 
-Analyzing the memory can help to identify the root cause of different problems, like for example why an application is crashing, but can also be used to identify sensitive data. This section describes how to check for sensitive data and disclosure of data in general within the process memory.
+Analyzing memory can help developers to identify root causes of different problems, like for example application crashes, etc. However, it can also be used to gain access to sensitive data. This section describes how to check for disclosure of data within the process' memory.
```
Sjord: > like for example application crashes, etc. 

This should be "such as application crashes". I think "like", "for example" and "etc" is a bit too much to indicate this is just an example.

> different problems

Shouldn't this be "several problems"?

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -1069,22 +1069,35 @@ setContentView(R.layout.activity_main);
...
+
+First, you need to identify which sensitive information is stored in memory. Basically, if you have a sensitive asset it's very likely that at some point it is loaded in memory. The objective is to verify that these info has as short exposure as possible.
```
Sjord: > these info

I think this should be "this information".

> has as short exposure as possible.

I think some word is missing here. Maybe it's better to make it "is exposed as briefly as possible."

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -1069,22 +1069,35 @@ setContentView(R.layout.activity_main);
...
 Once sensitive functions are identified, like decryption of data, the investigation of a memory dump might be beneficial in order to identify sensitive data like a key or the decrypted information itself.
 
 #### Static Analysis
 
-First, you need to identify which sensitive information is stored in memory. Then there are a few checks that must be executed:
+Before looking into the source code, it is beneficial to check documentation (if available) and identify application components so that you get the big picture of where a certain data might be exposed. For example, a sensitive data received from backend does not only exist in the final model object, but also might have multiple copies in the HTTP client, the XML parser, etc. Ideally you want all of these copies to be removed from memory as soon as possible.
```
Sjord: > a data

Data is a bit of a strange word because it is sometimes considered plural and singular. In any case, I would just use "data" here, not "a data".

Sjord:
I am not sure I fully understand what the attack vector is that we want to protect against by clearing memory. Isn't it possible to access a key in memory from a debugger, no matter how brief it is used? Is there another way that memory leaks?

sushi2k:
You can dump the memory even without a debugger on a non-jailbroken device, see [objection](https://github.com/sensepost/objection/) for iOS or fridump in general. 
The attack vector is of course quite sophisticated, as you need access to the device either physically or through malware to be able to exploit this. 

As Pece was writing:

> The objective is to verify that these info has as short exposure as possible. 

At some point sensitive information will be in memory, but you can still apply best practices to make this harder for an attacker to get in possession of the secrets or other sensitive data.

pmilosev:
@Sjord thanks for the syntax check.
Regarding your last comment - that's why I also added a paragraph suggesting that analysis should be done if some particular data has to be in memory in the first place. If my example of an encrypted data received from one server and transfered to another is not clear enough, maybe I should elaborate a little bit on it, or provide an alternative example.

pmilosev:
@Sjord @sushi2k @b-mueller 
I finished editing this section.
Please do a review and merge the pull request.

Whoever has deeper Android knowledge, please also verify that everything here makes sense.
I basically added my input on top of what was already present, but there are some things I'm not sure about (e.g. If Eclipse Memory analyzer really only provides access to static classes over the Dominator Tree, or if the statements about SecretKey are true).

I'll have some time to edit the content in the next days and then I'll have to fade alway for a while again :)

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -1069,166 +1069,244 @@ setContentView(R.layout.activity_main);
...
+First, you need to identify which sensitive information is stored in memory. Basically, if you have a sensitive asset it's very likely that at some point it is loaded in memory. The objective is to verify that this info is exposed as briefly as possible.
+
+To be able to investigate the memory of an application a memory dump needs to be created first. Alternatively it can be analyzed in real-time, e.g. over a debugger. No matter the approach, this is a very error prone process from a verification point of view, as what you will get in a certain dump is the data left by the functions that were executed. You might miss executing critical scenarios. Additionally, unless you know the footprint of the data you are looking for (either the exact value, or its format), it is quite easy not to identify it during analysis. For example, if the app performs encryption based on a randomly generated symmetric key, unless you get to know the value of the key by other means, it is very unlikely that you will be able to spot it in memory.
```
sushi2k: " a certain dump is the data left " -> " a certain dump **of** the data left "

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -1069,166 +1069,244 @@ setContentView(R.layout.activity_main);
...
+
+Therefore you are better off starting with static analysis.
 
 #### Static Analysis
 
-First, you need to identify which sensitive information is stored in memory. Then there are a few checks that must be executed:
+Before looking into the source code, it is beneficial to check documentation (if available) and identify application components so that you get the big picture of where certain data might be exposed. For example, sensitive data received from backend does not only exist in the final model object, but also might have multiple copies in the HTTP client, the XML parser, etc. Ideally you want all of these copies to be removed from memory as soon as possible.
```
sushi2k: -> "from the backend"

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -1069,166 +1069,244 @@ setContentView(R.layout.activity_main);
...
+In order to be able to analyze the dump in MAT you need to use the _hprof-conv_ platform tool, provided with the Android SDK. 
 
 ```bash
-./hprof-conv file.hprof file-converted.hprof
+./hprof-conv memory.hprof memory-mat.hprof
 ```
 
-By using MAT, more functions are available, like usage of the Object Query Language (OQL). OQL is an SQL-like language that can be used to make queries in the memory dump. Analysis should be done on the dominator tree as only this contains the variables/memory of static classes.
+MAT provides several different tools you can use to analyze the memory dump. For example, you can use the _Histogram_ to get an idea on how many objects have been captured from a certain type, or the _Thread Overview_ to see process' threads and their stack frames. Check the _Dominator Tree_ to learn about keep-alive dependencies between objects. (**TO BE CONFIRMED OR REMOVED:** The _Dominator Tree_ is also the only tool containing the information for static classes). You can use regular expressions to filter out the results in all of these tools.
```
sushi2k: Cannot find any evidence for that at the moment. But as far as I know only the dominator tree contains the static classes.

sushi2k:
Made a quick review. Looks good to me. Really good 👍

Document/0x05d-Testing-Data-Storage.md
pmilosev: hmmm OF doesn't looks right here:

... what you will get in a certain dump OF the data left ...
... what you will get in a certain dump IS the data left ...

WIth OF it sounds like an open question, while my intention here is to provide the answer ... so THE DATA LEFT is the answer to WHAT YOU WILL GET

Document/0x05d-Testing-Data-Storage.md
pmilosev: probably should be "from a backend" as we I'm talking in general here not referring to some particular backend.

Document/0x05d-Testing-Data-Storage.md
pmilosev: I'm not looking for evidence :)
I just never came across that info ... and in the one or two cases where I actually have done memory analysis I never had to look for static classes (assuming you mean static nested classes here, as Java does not support static top-level classes ... or did you mean top-level classes with only static methods ?)

Document/0x05d-Testing-Data-Storage.md
sushi2k: When I read it again, of doesn't make much sense. You are right. 
Keep it as it is :-)

Document/0x05d-Testing-Data-Storage.md
sushi2k: I think someone told me this during a conference a while ago, that in the dominator tree you can find information about the static classes and methods. I would say delete this sentence, as during analysis you are looking for certain patterns, regardless if they are from a static class or not.

PR #678, opened 2017-07-19
Add crypto key management Android best practices and testing guides 

DarioI:


Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
...
+
+The most secure way of handling key material, is simply never storing them on the filesystem. This means that the user should be prompted to input a passphrase every time the application needs to perform a cryptographic operation. Although this is not the ideal implementation from a user experience point of view, it is however the most secure way of handling key material. The reason is because key material will only be available in an array in memory while it is being used. Once the key is not needed anymore, the array can be zeroed out. This minimizes the attack window as good as possible. No key material touches the filesystem and no passphrase is stored. The encryption key can be generated from the passphrase by using the Password Based Key Derivation Function version 2 (PBKDFv2). This cryptographic protocol is designed to generate secure and non brute-forceable keys.
```
commjoen: Next, it would be a good idea to give an example of PBKDFv2 because people will try to do it themselves, making mistakes in the process...

Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
...
+
+A sligthly less secure way of storing encryption keys, is in the SharedPreferences of Android. When [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences.html"Android SharedPreference API") are initialized in [MODE_PRIVATE](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE, the file is only readable by the application that created it. However, on rooted devices any other application with root access can simply read the SharedPreference file of other apps, it does not matter whether MODE_PRIVATE has been used or not. This is not the case for the KeyStore. Since KeyStore access is managed on kernel level, which needs considerably more work and skill to bypass without the KeyStore clearing or destroying the keys. 
+
```
commjoen: If this is used, then at least consider the BC (UBER) keystore which adds encryption on top of the filesystem :).

Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
...
+- `Cipher`
+- `Mac`
+- `MessageDigest`
+- `Signature`
+- `KeyStore`
+- `Key`, `PrivateKey`, `PublicKey`, `SecretKey`
+- And a few others in the `java.security.*` and `javax.crypto.*` packages.
+
+Trace and assess how key material is generated and stored before or after it is used.
```
commjoen: It would be nice to give a bit more details on how one is using it wrong in the static analysis part

Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
 
 ##### Info
 - [#nelenkov] - N. Elenkov, Android Security Internals, No Starch Press, 2014, Chapter 5.
+
+
+### Testing Key Management
+
+#### Overview
+
+Symmetric cryptography provides confidentiality and integrity of data because it ensures one basic cryptographic principle. It is based on the fact that a given ciphertext can only, in any circumstance, be decrypted when providing the original encryption key. The security problem is thereby shifted to securing the key instead of the content that is now securely encrypted. Asymmetric cryptography solves this problem by introducing the concept of a private and public key pair. The public key can be distributed freely, the private key is kept secret.
```
Sjord: > Symmetric cryptography provides confidentiality and integrity of data

Encrypting something does not necessarily ensures the integrity of the data. With a [bit-flipping attack](https://en.wikipedia.org/wiki/Bit-flipping_attack), for example, it is possible to tamper with encrypted data.

Still, I think this sentence is acceptable as it is, because "symmetric cryptography" may also include MAC functionality and some encryption algorithms also provide message integrity.

Document/0x05e-Testing-Cryptography.md
Sjord: > The most secure way of handling key material, is simply never storing them on the filesystem.

This should be "storing *it* on the filesystem."

Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
...
+
+Now, it is clear that regularly prompting the user for its passphrase is not something that works for every application. In that case make sure you use the [Android KeyStore API]("Android KeyStore API"). This API has been specifically developed to provide a secure storage for key material. Only your application has access to the keys that it generates. Starting from Android 6.0 it is also enforced that the KeyStore is hardware-backed. This means a dedicated cryptography chip or trusted platform module (TPM) is being used to secure the key material.
+
```
Sjord: > [Android KeyStore API]("Android KeyStore API")

I think you are missing a link here.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
...
+A sligthly less secure way of storing encryption keys, is in the SharedPreferences of Android. When [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences.html"Android SharedPreference API") are initialized in [MODE_PRIVATE](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE, the file is only readable by the application that created it. However, on rooted devices any other application with root access can simply read the SharedPreference file of other apps, it does not matter whether MODE_PRIVATE has been used or not. This is not the case for the KeyStore. Since KeyStore access is managed on kernel level, which needs considerably more work and skill to bypass without the KeyStore clearing or destroying the keys.
```
Sjord: Please check the syntax of both links. On eis missing a space and the other is missing a closing parenthesis.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -233,3 +233,52 @@ public static void main (String args[]) {
...
+- V3.1: "The app does not rely on symmetric cryptography with hardcoded keys as a sole method of encryption."
+- V3.5: "The app doesn't reuse the same cryptographic key for multiple purposes."
+##### OWASP Mobile Top 10 2016
+- M6 - Broken Cryptography
+
+##### CWE
+- CWE-321: Use of Hard-coded Cryptographic Key
+
+##### Info
+- [#nelenkov] - N. Elenkov, Android Security Internals, No Starch Press, 2014, Chapter 5.
```
Sjord: Maybe this should go under references.

Document/0x05e-Testing-Cryptography.md
commjoen: HMACing can use symmetric crypto indeed to provide integrity. Another option is to use GCM mode :).

Document/0x05e-Testing-Cryptography.md
Sjord: Please move this line directory under "References" and remove the "Info" section.

Document/0x05e-Testing-Cryptography.md
DarioI: Any recommendations for the BC library that does not properly handle arrays with key material? Adding the example in my next commit.

Document/0x05e-Testing-Cryptography.md
DarioI: I do not see how this will improve the security. Where are you going to store the key of the encrypted filesystem (assuming you mean a sort of encrypted sharedpreferences)? The only safer place I see is using the KeyStore, but why are they using the SharedPreferences in the first place to store the other encryption keys?

Document/0x05e-Testing-Cryptography.md
DarioI: I did not go into to much detail on the different cryptographic algorithms and protocols, since the writing guide says to keep it to point and informative..

Document/0x05e-Testing-Cryptography.md
Sjord: We already mention that BouncyCastle doesn't clean up properly in [0x05d](https://github.com/OWASP/owasp-mstg/blob/master/Document/0x05d-Testing-Data-Storage.md):
> Lastly, some ciphers do not properly clean up their byte-arrays. For instance, the AES Cipher in BouncyCastle does not always clean up its latest working key.

DarioI:
Normally I should have all the changes except for the removal of the Info line, this has already been done in another commit and needs to be resolved during merging. Looking forward to the feedback.

Document/0x05e-Testing-Cryptography.md
b-mueller: IMHO the core problem that asymmetric crypto solves is *sharing* the key, not *securing* it.

DarioI: True, good catch! That was actually what I meant but got lost during writing it :-)

Document/0x05e-Testing-Cryptography.md
```diff
@@ -232,3 +232,86 @@ public static void main (String args[]) {
...
+public static SecretKey generateStrongAESKey(char[] password, int keyLength)
+{
+    //Initiliaze objects and variables for later use
+    int iterationCount = 10000;
+    int saltLength     = keyLength / 8;
+    SecureRandom random = new SecureRandom();
+    
+    //Generate the salt
+    byte[] salt = new byte[saltLength];
+    randomb.nextBytes(salt);
```
Sjord: It seems to me that if you use a random salt, the generated key will be different every time. This means that this function is unusable to derive a key from a password, because it will be different every time.

DarioI: Yes, you are right, the salt needs to remain the same when you want to derive the same encryption key for the same password, that's why I specified this in the text below that code snippet. Hozever, I wanted to illustrate that you can generate the salt and store it at the moment you generate the key for the first time. Instead of using a hardcoded salt.

b-mueller: What's the advantage of using the salt in this case?

DarioI: A salt is usually used to counter a pre-computated attack of encyption keys or hashes also known as a rainbow table attack. An attacker pre-computes all the different possibilities and stores them in a big table. These tables can easily reach several gigabytes in size. When an attacker then wants to bruteforce  a key or hash he simply iterates over the table of pre-computated hashses or  keys and tries them one after one. He does not need to do the work of generating or computing the keys/hashes anymore. By adding a salt to the key or hash generation phase you basically render these rainbow tables useless. An attacker will need to use the hash and re-compute all the different hashes/keys in a brute-force attack. Because the PBKDFv2 is specifically designed to slow down the generation process (by increasing the iteration count) this will significantly increase the effort for a brute force attack.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -232,3 +232,86 @@ public static void main (String args[]) {
...
+
+The most secure way of handling key material, is simply never storing it on the filesystem. This means that the user should be prompted to input a passphrase every time the application needs to perform a cryptographic operation. Although this is not the ideal implementation from a user experience point of view, it is however the most secure way of handling key material. The reason is because key material will only be available in an array in memory while it is being used. Once the key is not needed anymore, the array can be zeroed out. This minimizes the attack window as good as possible. No key material touches the filesystem and no passphrase is stored.  However, note that some ciphers do not properly clean up their byte-arrays. For instance, the AES Cipher in BouncyCastle does not always clean up its latest working key.
```
Sjord: > The reason is because

It is either "The reason is" or "That is because".

Document/0x05e-Testing-Cryptography.md
Sjord: This should be `random.nextBytes` instead of `randomb.nextBytes`.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -232,3 +232,86 @@ public static void main (String args[]) {
...
+
+The last two options are to use hardcoded encryption keys in the source code and storing generated keys in public places like ```/sdcard/```. Obviously, hardcoded encryption keys are not the way to go. This means every instance of the application uses the same encryption key. An attacker needs only to do the work once, to extract the key from the source code . Consequrently, he can decrypt any other data that he can obtain and that was encrypted by the application. Lastly, storing encryption keys publicly also is highly discouraged as other applications can have permission to read the public partition and steal the keys.
```
Sjord: > from the source code . Consequrently

Please remove the space before the period. Also, it probably be "Consequently".

Document/0x05e-Testing-Cryptography.md
```diff
@@ -232,3 +232,86 @@ public static void main (String args[]) {
...
+
+- `Cipher`
+- `Mac`
+- `MessageDigest`
+- `Signature`
+- `KeyStore`
+- `Key`, `PrivateKey`, `PublicKey`, `SecretKeySpec`
+- And a few others in the `java.security.*` and `javax.crypto.*` packages.
+
+As an example we illustrate how to locate the use of a hardcoded encryption key. First disassmble the DEX bytecode to a collection of Smali bytecode files using ```Baksmali```.
```
Sjord: > First disassmble

disassemble

PR #689, opened 2017-07-25
Update style_guide.md

caitlinandrews:
Added information about contractions, serial commas in lists, and slang


b-mueller:
Thanks @caitlinandrews

PR #727, opened 2017-08-05
Mini procedure to identify detection of rooted device

righettod:
Hello,

I have added a mini procedure in order to help to statically identify code that can be used to detect if the device is rooted.

Thanks in advance.


righettod:
Thank you for the merging.

PR #736, opened 2017-08-14
Update 0x06d-Testing-Data-Storage.md

### LINKED ISSUE #732: 0x06d-Testing-Data-Storage - Testing for Sensitive Data in the Clipboard
I suggest to add the following to the Overview section within Testing for Sensitive Data in the Clipboard:

Before iOS 9, a malicious app might monitor the pasteboard in the background while periodically retrieving `[UIPasteboard generalPasteboard].string`. As of iOS 9, the access to the pasteboard content is only allowed to apps in the foreground.

However, just a suggestion. Would be nice to see in the guide.

allRiceOnMe:
This commit fixes #732


sushi2k:
Thanks @allRiceOnMe

PR #746, opened 2017-08-19
0x05a-Platform-Overview line 130 - 252

AnnaSzk:
minor linguistic updates

### DIFF
--- a/Document/0x05a-Platform-Overview.md
+++ b/Document/0x05a-Platform-Overview.md
@@ -129,7 +129,7 @@ drwxrwx--x u0_a65   u0_a65            2016-01-10 09:44 shared_prefs
 
 #### APK Structure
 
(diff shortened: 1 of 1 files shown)


Document/0x05a-Platform-Overview.md
```diff
@@ -239,7 +240,7 @@ drwx------  4 u0_a97              u0_a97              4096 2017-01-18 14:27 com.
 drwx------  6 u0_a120             u0_a120             4096 2017-01-19 12:54 com.android.chrome
 ```
 
-However, if two apps are signed with the same certificate and explicitly share the same user ID (by including the _sharedUserId_ in their _AndroidManifest.xml_) they can access each others data directory. See the following example on how this is achieved in the Nfc app:
+However, if two apps are signed with the same certificate and explicitly share the same user ID (by including the _sharedUserId_ in their _AndroidManifest.xml_) they can access each other’s data directory. See the following example on how this is achieved in the Nfc app:
```
Sjord: 👍

PR #751, opened 2017-08-21
Fix Typo - 0x04e-Testing-Authentication-and-Session-Management

TommyJ1994:
"The app will generate a public/private key pair when the user `signs` (was sigs) up, ...."


Sjord:
Thanks @TommyJ1994

PR #753, opened 2017-08-22
Fix typo in Platform Overview - Zygote Section

TommyJ1994:
In the last sentence : "it forks a new process which then loads and executes `the the` app-specific code."

->  "it forks a new process which then loads and executes `the` app-specific code."


Sjord:
👍

PR #828, opened 2017-10-06
0x4b - Remove redundant content

romualdszkudlarek:
The removed section was appearing twice in the chapter. Therefore, one of the 2 sections has been removed.


b-mueller:
Thanks Romuald!

PR #829, opened 2017-10-08
Update style_guide.md

hlhodges:
Added section "In-text keywords" to the end of this guide. Please verify: Should all of the following be put in backticks when not in a code block?

function names
method names
class names
commands
flags (true/false)
command line utility names
directory names
all file names? The config file names I see are formatted in all capital letters; that seems standard.


b-mueller:
Thanks Heaven!

PR #850, opened 2017-11-12
Add broadcast info

B3nac:
Adding commands that can be used to test vulnerable broadcast receivers.


b-mueller:
Thank you @B3nac

PR #856, opened 2017-11-25
Adding an XSS example via start ContentActivity 

B3nac:
Let me know if I need to make any further changes regarding format. 🎉


b-mueller:
thanks @b3bac!

PR #857, opened 2017-11-28
 App sandbox dirs updates and IPA installer references 

raulsiles:
App sandbox dirs updates and IPA installer references consistency.


sushi2k:
Thanks Raul.

PR #864, opened 2018-01-02
SQL injection examples.

B3nac:
Please let me know if this checks out. Happy New Year!


Document/0x04h-Testing-Code-Quality.md
```diff
@@ -44,6 +44,47 @@ SELECT * FROM users WHERE username='1' OR '1' = '1' AND Password='1' OR '1' = '1
 
 Because the condition `'1' = '1'` always evaluates as true, this query return all records in the database, causing the login function to return "true" even though no valid user account was entered.
 
+Here is another example of `'1' = '1'` that dumps all tables in a database via adb from [Ostorlab Blog](http://blog.ostorlab.co/2016/03/android-sql-and-contentproviders-or-why.html).
```
Sjord: I don't think this is an 1=1 payload example at all. In the previous example, we inject an extra SQL condition, `or 1=1`, so that all users are returned. In your example, the payload is:

    _id/**/limit/**/\(select/**/1/**/from/**/sqlite_master/**/where/**/1=1\)

This is another form of SQL injection, and 1=1 is just used in the query that the attacker wants to run. In fact, it seems to me that this works just as well:

    _id/**/limit/**/\(select/**/1/**/from/**/sqlite_master\)

B3nac:
Hi Sjord,

I should have changed the wording to another true  condition because it's not exactly the same. I can change the description after work.

Document/0x04h-Testing-Code-Quality.md
sushi2k: Agree with Sjord. The root cause of this SQL Injection is in the sort parameter. Can you please rephrase @B3nac ?

B3nac: @Sjord @sushi2k Thank you for the feedback, I'm on it. :]

B3nac: @Sjord @sushi2k I think those commits cover everything. Please let me know if that checks out. I apologize for the delay, I was busy moving a crossed the U.S. and college classes started back up.

sushi2k:
Thanks for PR

PR #869, opened 2018-02-05
iOS Needle Documentation

HenryHoggard:
I've begun documenting how Needle can be used to test iOS applications. More to come just committing the sections I've done so far. Also added Keychain data persistence issue.


sushi2k:
Thanks for adding. Just merged it

PR #886, opened 2018-05-03
Updated HelloWord-JNI.apk URL

Chan9390:
`OMTG-Files` directory no more exists. Also the previous URL uses `/blob/master/` which give HTML output even if the apk exists.

**Changes:**
- Changed the URL to `https://github.com/OWASP/owasp-mstg/raw/master/Samples/Android/01_HelloWorld-JNI/HelloWord-JNI.apk` so that wget downloads the apk file


sushi2k:
Thanks

PR #887, opened 2018-05-03
Updated mistake

Chan9390:
**Changes:**
- Changed `&lt;application&gt;` to `<application>` because `&lt;` and `&gt;` doesnt render inside markdown code
- Changed quote to `


sushi2k:
Thanks for fixing

PR #888, opened 2018-05-03
Added frida-server download command

Chan9390:
The docs shows two different commands: one to find the version and other one to download the frida-server binary after manually updating the version in the url. (As of now the latest frida version downloaded using pypi is: 10.8.2)

I added single bash command `wget https://github.com/frida/frida/releases/download/$(frida --version)/frida-server-$(frida --version)-android-arm.xz` which eases the download of the right binary in single step.


sushi2k:
Thanks. That makes it easier!

PR #890, opened 2018-05-22
Update to the 0x06d-Testing-Data-Storage.md

hduarte:
Applied a few text fixes and changed the order of a few points to make things clear.


sushi2k:
Thanks!

PR #933, opened 2018-07-23
Added Russian translation of iOS related chapters(0x06).

kysokzla:
I've finished translation & reviewing of OWASP MSTG iOS related chapters. 
There is a small possibility that translation of general chapters and Android related would be added later. Also i'm contributing MASVS Russian translation.


sushi2k:
Thanks for the PR. Amazing work :-)

PR #998, opened 2018-09-17
Fix Slack URL

TheDauntless:
Small fix for the Slack URL, since the herokuapp now links to the juiceshop project. This is the URL that is shared in the main owasp slack channel, but it will expire after 2000 new joiners.


commjoen:
Thanks!

PR #1030, opened 2018-10-11
Add information to allow application installation on a non-iPad device

righettod:
Hello,

The PR add information to handle the case in which an application require an iPad and the auditor have only iPhone or iPod devices.

Checklist state:

Thanks you very much in advance.

### DIFF
--- a/Document/0x06b-Basic-Security-Testing.md
+++ b/Document/0x06b-Basic-Security-Testing.md
@@ -573,3 +573,26 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
 ```shell
 ip.addr == 192.168.1.1 && http
 ```
+
+### Allow application installation on a non-iPad device
+
+Sometimes an application can require to be used on a iPad device. If you only have iPhone or iPod touch devices then you can force the application to accept to be installed and used on these kinds of devices by changing the value of the property **UIDeviceFamily** to the value **1** in the **Info.plist** file. 
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+
+  <key>UIDeviceFamily</key>
+  <array>
+    <integer>1</integer>
+  </array>
+
+</dict>
+</plist>  
+```
+
+It is important to note that changing this value will break the original signature of the IPA file so you need to re-sign the IPA, after the update, in order to install it on a device on which the signature validation has not been disabled.
+
+Possible values for the property [UIDeviceFamily](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/iPhoneOSKeys.html#//apple_ref/doc/uid/TP40009252-SW11 "UIDeviceFamily property").
--- a/Document/0x06b-Basic-Security-Testing.md
+++ b/Document/0x06b-Basic-Security-Testing.md
@@ -574,7 +574,7 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
 ip.addr == 192.168.1.1 && http
 ```
 
-### Allow application installation on an non-iPad device
+### Allow Application Installation on an Non-Ipad Device
 
 Sometimes an application can require to be used on an iPad device. If you only have iPhone or iPod touch devices then you can force the application to accept to be installed and used on these kinds of devices. You can do this by changing the value of the property **UIDeviceFamily** to the value **1** in the **Info.plist** file. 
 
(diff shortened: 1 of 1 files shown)


Document/0x06b-Basic-Security-Testing.md
```diff
@@ -573,3 +573,26 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
 ```shell
 ip.addr == 192.168.1.1 && http
 ```
+
+### Allow application installation on a non-iPad device
+
+Sometimes an application can require to be used on a iPad device. If you only have iPhone or iPod touch devices then you can force the application to accept to be installed and used on these kinds of devices by changing the value of the property **UIDeviceFamily** to the value **1** in the **Info.plist** file.
```
commjoen: 1. can you please change "on a iPad device" to "on an iPad device"?
2. can you please split sentence into `and used on these kinds of devices. You can do this by changing`

Document/0x06b-Basic-Security-Testing.md
```diff
@@ -573,3 +573,26 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
...
+    <integer>1</integer>
+  </array>
+
+</dict>
+</plist>  
+```
+
+It is important to note that changing this value will break the original signature of the IPA file so you need to re-sign the IPA, after the update, in order to install it on a device on which the signature validation has not been disabled.
+
+Possible values for the property [UIDeviceFamily](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/iPhoneOSKeys.html#//apple_ref/doc/uid/TP40009252-SW11 "UIDeviceFamily property").
```
commjoen: Can you add this to the references section as well?

Document/0x06b-Basic-Security-Testing.md
```diff
@@ -573,3 +573,26 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
...
+  <key>UIDeviceFamily</key>
+  <array>
+    <integer>1</integer>
+  </array>
+
+</dict>
+</plist>  
+```
+
+It is important to note that changing this value will break the original signature of the IPA file so you need to re-sign the IPA, after the update, in order to install it on a device on which the signature validation has not been disabled.
```
commjoen: Please add a note that this might not work if the app requires capabilities which are not on the iPad.

Document/0x06b-Basic-Security-Testing.md
righettod: Hi @commjoen, 

**references section** is a special document in MSTG because I do not find it in the iOS documents section ?

Thanks in advance for your help.

Document/0x06b-Basic-Security-Testing.md
righettod: For sure.

Document/0x06b-Basic-Security-Testing.md
```diff
@@ -573,3 +573,28 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
 ```shell
 ip.addr == 192.168.1.1 && http
 ```
+
+### Allow application installation on an non-iPad device
```
sushi2k: Can you please adjust the title according to the style guide? 

https://github.com/OWASP/owasp-mstg/blob/master/style_guide.md#title-capitalization

Document/0x06b-Basic-Security-Testing.md
sushi2k: This section does not have a refence sections, those are usually at the actual chapters with test cases. So just leave the link here. Thanks

Document/0x06b-Basic-Security-Testing.md
sushi2k: BTW: Thanks for the content, that is really useful!

righettod: It's done, i have changed it to **Allow Application Installation on an Non-Ipad Device**.

I have used the site https://capitalizemytitle.com using `Chicago` style.

The site http://www.titlecapitalization.com/ do not respond:
```
Firefox can't establish a connection to the server at www.titlecapitalization.com.
```

Document/0x06b-Basic-Security-Testing.md
```diff
@@ -573,3 +573,28 @@ You can remotely sniff all traffic in real-time on iOS by [creating a Remote Vir
...
+    <integer>1</integer>
+  </array>
+
+</dict>
+</plist>  
+```
+
+It is important to note that changing this value will break the original signature of the IPA file so you need to re-sign the IPA, after the update, in order to install it on a device on which the signature validation has not been disabled. 
+
+This bypass might not work if the application requires capabilities that are specific to iPad device.
```
commjoen: Maybe we can change this into:
`This bypass might not work if the application requires capabilities that are specific to modern iPads while your iphone or iPod is a bit older` and then it should be fine!

righettod:
Thanks you for the merging.

PR #1058, opened 2018-10-27
Fix markdown for code chunks

joscandreu:
- Added the proper language syntax to code chunks to get it properly highlighted.
- Fixed some inconsistencies in code indentation.
- Added some $ at the beginning of console commands.

There is more work pending around this. I think I will open an issue to talk about it and see if there is already a standard in this project, e.g. should all console commands start with $ to differentiate the input command from output? or max line length in code chunks...


commjoen:
Thanks!

sushi2k:
Thanks for the PR @joscandreu !

We even have this in our style guide:

https://github.com/OWASP/owasp-mstg/blob/master/style_guide.md#code-and-shell-commands

PR #1062, opened 2018-10-30
Update 0x08-Testing-Tools.md

dweinstein:


anantshri:
Why are we linking to company home page instead of specific tools page?

commjoen:
Good question: can  you file an issue?

PR #1129, opened 2019-01-31
update Frida detection on Android

sushi2k:

This PR covers issue #1130.


Document/0x05j-Testing-Resiliency-Against-Reverse-Engineering.md
```diff
@@ -785,31 +785,55 @@ You can detect popular reverse engineering tools that have been installed in an
...
 ```
 
-This works if Frida is run in its default configuration. Perhaps it's also enough to stump some script kiddies during their first steps in reverse engineering. It can, however, be easily bypassed by renaming the frida-server binary, so we should find a better method.
+Starting with Android Nougat (API Level 24) the `ps` command will only return processes started by the user itself. When executing `ps` it will read the information from `/proc` and it's not possible to access information that belongs to other user ids.
+
+![Executing ps on Android Lollipop](Images/Chapters/0x05j/Android_Lollipop_ps.png)
+
+![Executing ps on Android Nougat](Images/Chapters/0x05j/Android_Nougat_ps.png)
+
+Even if the process name could easily be detected, this would only work if Frida is run in its default configuration. Perhaps it's also enough to stump some script kiddies during their first steps in reverse engineering. It can, however, be easily bypassed by renaming the frida-server binary. So because of this and the technical limitations of querying the process names in recent Android versions, we should find a better method.
```
commjoen: So because of this and the technical limitations:
Therefore, we need to look for a better method of querying the process names in recent android versions.

commjoen: Additionally: you know that there will be a gadget in your patched runtime. This means that the app has been resigned with either the same cert and a different signature or with another cert with a different signature. Can you refer to integrity checking, or update the write up about that given v2/v3 signign scheme in android?
Because that way you can see that your binary has been modified.

sushi2k: I dont know of any to be honest. If it's a non-rooted device it seems you cannot query for  process names, only the one that matches with your uid.

Document/0x05j-Testing-Resiliency-Against-Reverse-Engineering.md
```diff
@@ -785,31 +785,55 @@ You can detect popular reverse engineering tools that have been installed in an
...
-  }
-  return returnValue;
+
+    return returnValue;
 }
+
 ```
 
-This works if Frida is run in its default configuration. Perhaps it's also enough to stump some script kiddies during their first steps in reverse engineering. It can, however, be easily bypassed by renaming the frida-server binary, so we should find a better method.
+Starting with Android Nougat (API Level 24) the `ps` command will only return processes started by the user itself. When executing `ps` it will read the information from `/proc` and it's not possible to access information that belongs to other user ids.
```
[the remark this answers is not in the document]
sushi2k: let me update that tomorrow. Thanks

Document/0x05j-Testing-Resiliency-Against-Reverse-Engineering.md
sushi2k: I was already thinking about integrity checking of the APK to verify if the APK was tampered . through repackaging. let me check if this is already described in the reverse engineering part.

commjoen: ah, then let's change it to:"
Therefore, we need to look for a better methods _than_ querying the process names in recent android versions.

Document/0x05j-Testing-Resiliency-Against-Reverse-Engineering.md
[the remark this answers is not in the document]
sushi2k: done

Document/0x05j-Testing-Resiliency-Against-Reverse-Engineering.md
sushi2k: done

sushi2k: I made a very simple check. Let me know  what you think

Issue #956, opened 2018-08-06
Android Oreo update: changes in permissions


commjoen:
The permissions structure is updated again in Android Oreo: https://developer.android.com/about/versions/oreo/android-8.0-changes

TheDauntless:
@commjoen Close please

PR #1176, opened 2019-04-15
Fix for #956

TheDauntless:

- [ ] Your contribution is written in the 2nd person (e.g. you)
- [ ] Your contribution is written in an active present form for as much as possible.
- [ ] Your contribution has proper formatted markdown and/or code
- [ ] Any references to website have been formatted as [TEXT](URL “NAME”)
- [ ] You verified/tested the effectiveness of your contribution (e.g.: is the code really an effective remediation? Please verify it works!)

This PR covers issue #956 

Most of it was already in there actually.


TheDauntless:
Apparently we also don't have anything in the style guide for referencing API levels. I feel this should also be consistent.

sushi2k:
That is very true. Let's stick to the way Android is also using it in it's documentation https://developer.android.com/studio/releases/platforms 

Android 9 (API level 28)

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -225,7 +225,7 @@ public void onRequestPermissionsResult(int requestCode, //requestCode is what yo
 }
 
 ```
-Permissions should be explicitly requested for every needed permission, even if a similar permission from the same group may change in the future. Also permissions may be granted without user approval automatically.
+Permissions should be explicitly requested for every needed permission, even if a similar permission from the same group has already been requested. For applications targeting SDK 25 (Android 7.1) and older, Android will automatically give an application all the permissions from a permission group, if the user grants one of the requested permissions of that group. Starting with SDK 26 (Android 8.0), permissions will still automatically be granted if a user has already granted a permission from the same permission group, but the application still needs to explicitly request the permission. In this case, the `onRequestPermissionsResult` handler will automatically be triggered without any user interaction.
```
sushi2k: Let's use this format with API Level instead of SDK and put the SDK into brackets:

Android 9 (API level 28)

TheDauntless: Done.

PR #1224, opened 2019-05-29
Updated key areas section

Chan9390:
Changed "mobile app penetration testing tools" to "mobile app penetration testers".


sushi2k:
Good spot! Thanks @Chan9390

PR #1343, opened 2019-07-16
Drop Burp-Non-HTTP-Extension

Jnchi:
NoPE Proxy doesn't appear to be maintained anymore.

Bug: https://github.com/summitt/Burp-Non-HTTP-Extension/issues/35

### DIFF
--- a/Document/0x04f-Testing-Network-Communication.md
+++ b/Document/0x04f-Testing-Network-Communication.md
@@ -22,7 +22,6 @@ Using a proxy breaks SSL certificate verification and the app will usually fail 
 
 Interception proxies such as Burp and OWASP ZAP won't show non-HTTP traffic, because they aren't capable of decoding it properly by default. There are, however, Burp plugins available such as:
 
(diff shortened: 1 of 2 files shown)


sushi2k:
Hi @Jnchi. Thanks for the PR! 
Even though that it wasn’t updated it’s one of the rare Burp plugins you can use when the app is not using HTTP and it’s still doing it’s job (was using it last year for a Pentest).
It should also still work with the recent stable 2.0 version of Burp, as the Extension APIs didn’t change (please correct me if I am wrong). 
So you can still use it and therefore I would say we should keep it to at lest give some guidance what to do if you need to assess an app that is not using HTTP.

Jnchi:
Let me get back to you on this.

sushi2k:
If you have any other tool or plugin that we can use to replace NoPe, please feel free to make another PR. It's just not good if we get rid of something that is still working without having a proper replacement and especially in this area (testing non-http communication) there is really not much. 

I will close this PR for now and feel free to re-open or create another PR. Thanks

PR #1417, opened 2019-08-20
Remove password rules and checks that contradict the cited cheat sheet.

oliver-gramberg:
The OWASP Authentication Cheat Sheet cited for details on checking passwords and found at https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html further refers to https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v21-password-security-requirements, which explicitly states, "2.1.9 ... There should be **no** requirement for upper or lower case or numbers or special characters." (emphasis mine.)

It also does not (no longer) have a rule that forbids repeat characters.

This seems to be in line with the 2019 NIST password recommendation contained in  https://pages.nist.gov/800-63-3/sp800-63b.html


commjoen:
Thank you for your PR! I will have a look soon at it.

PR #1481, opened 2019-10-03
introduce FileDP tool for Data Protection Class verification

abjurato:
```- [ ] You verified/tested the effectiveness of your contribution```
👆 This part is especially funny because Needle framework is pretty abandoned and supports only iOS 8, 9 and 10 and most modules support only 32 bit devices. Rather then enhancing `Dynamic Analyse with Needle` section we'd better consider replacing it with dynamic analyse with separate tools that once were included as modules into Needle.

This PR contains only a mention that Data protection class of a file can be verified during dynamic analysis. Hope interested reader will find her way to run `FileDP` on modern devices with modern jailbreaks.

### DIFF
--- a/Document/0x06d-Testing-Data-Storage.md
+++ b/Document/0x06d-Testing-Data-Storage.md
@@ -332,6 +332,14 @@ To use Needle to read the Keychain, execute the following command:
 [needle][keychain_dump] > run
 ```  
 
+###### Validate the Data Protection Classes
+You can use FileDP module in order to verify Data protection class of the file or directory:
+
+``` shell
+[needle] > use storage/data/FileDP
+[needle][FileDP] > run
+```
+
 ###### Searching for Binary Cookies
 
 iOS applications often store binary cookie files in the application sandbox. Cookies are binary files containing cookie data for application WebViews. You can use Needle to convert these files to a readable format and inspect the data. Use the following Needle module, which searches for binary cookie files stored in the application container, lists their data protection values, and gives the user the options to inspect or download the file:


cpholguera:
Hi @abjurato you definitely have a point there ^^ We are aware of this and have an issue open, please see: https://github.com/OWASP/owasp-mstg/issues/1409

For instance,  what FileDP does is already covered (and comes _for free_) whenever you run `ls` in objection. That should be the preferred way of checking the Data Protection classes in deed. Nowadays and as we recommend in the guide, any tester would have installed Frida on their device. So no additional installations are required. Just using objection will be enough for a task like this one.

If you feel like addressing this topic, we're more than happy about any help on this :) I definitely like your suggestion: "enhancing `Dynamic Analysis with Needle` section we'd better consider replacing it with Dynamic Analysis with separate tools that once were included as modules into Needle."

cpholguera:
I still have two questions @abjurato  ^^
- did the FileDP module worked for you? Do you have some example output from running it from needle?
- did you have to install it additionally to Needle or did it just work?

Thank you in advance!

abjurato:
@cpholguera I've found source of `FileDP` in author's repository https://github.com/satishb3/FileDp-Source , Needle's one is 32bit only

kysokzla:
@abjurato  we have run needle from very old ios version to 13.0. It's just a convinient wrapper for making calls to real binaries that gets the job done. It's just python2. Have u seen the code? What it do under the hood. It just calls binaries if ur jb solution do not include arm64 or arm64e binaries i have to compile it and replace in the right folder and u are good to go.

PR #1639, opened 2020-01-22
0x05c Android library injection

### LINKED ISSUE #1362: 5c - Create Library Injection
Create Tampering and RI -> Patching ... -> Library Injection

before end of "Patching ...", before "Dynamic instrumentation"

```
##### Library Injection
```

This section will talk about how to perform Library Injection on Android. This is used for example to inject Frida gadget but can be used on other cases and using several techniques: e.g. modifying the source code, patching a binary or its libraries, using a dynamic linker feature like LD_PRELOAD.

su-vikas:
This PR closes #1362 

Added content for android library injection.


su-vikas:
I didn't include the suggestion made in the ticket by @commjoen. I made this section generic, and didn't focus on frida. Let me know if this can still be included.

cpholguera:
Perfect, thank you @su-vikas I'll take a look at it soon ;)

Document/0x05c-Reverse-Engineering-and-Tampering.md
```diff
@@ -1332,6 +1332,62 @@ The following approach can be used in order to patch the JavaScript file:
...
...e it has its own limitations. For example, we have a 3rd party application for which we don't have the source code and only works locally. We want to modify this application so that it is accessible over the network as well. One possible way is to patch the application's APIs and introduce networking code, but this approach may not be practical for a large application. Another shortcoming being the patching performed is not reusable for any other executable, i.e patching performed is unique to each application. As an alternative approach, if we can separately develop and compile a library which provides the networking logic and inject this into the application, not only this will be time efficient, but also the library will be reusable. This paradigm of introducing a separately compiled code in a form of a library in a process memory is often referred to as library injection. In Windows world this technique has been in use for a long time, and is referred as *DLL injection*. It is commonly used to modify and bypass anti-cheat mechanisms in games. On Android and iOS, most common example is Frida, under-the-hood it is performing a library injection to provide all the functionalities.
```
cpholguera: Split in 2 paragraphs: first motivation/limitations, second intro/description. The Frida example use case is a bit more accurate.
```suggestion
In the previous section we learned about patching application code to assist in our analysis, but this approach has several limitations. For instance, you'd like to log everything that's being sent over the network without having to perform a MITM attack. For this you'd have to patch all possible calls to the network APIs, which can quickly become impracticable when dealing with large applications. In addition, the fact that patching is unique to each application can be also considered a shortcoming, as this code cannot be easily reused.

Using library injection you can develop reusable libraries and inject them to different applications, effectively making them behave differently without having to modify their original source code. This is known as DLL injection on Windows (broadly used to modify and bypass anti-cheat mechanisms in games), `LD_PRELOAD` on Linux and `DYLD_INSERT_LIBRARIES` on macOS. On Android and iOS, a common example is using the Frida Gadget whenever Frida's so-called [Injected mode](https://frida.re/docs/modes/#injected "Frida Injected Mode") of operation isn’t suitable (i.e. you cannot run frida-server on the target device). In this situation, you can [inject the Gadget](https://frida.re/docs/gadget/ "Frida Gadget") library by using the same methods you're going to learn in this section.
```

su-vikas:
Fixed as per suggestions, let me know how it looks now.

Document/0x05c-Reverse-Engineering-and-Tampering.md
```diff
@@ -1332,6 +1332,66 @@ The following approach can be used in order to patch the JavaScript file:
...
+
+###### LD_PRELOAD
+
+Above we looked into techniques which require some kind of modification of the application's code. A library can also be injected into a process using functionalities offered by the loader of the operating system. On Android, which is a Linux based OS, you can load an additional library by setting the `LD_PRELOAD` environment variable.
+
+The symbols loaded from the library passed using LD_PRELOAD always gets the precedence, i.e they are searched first by the loader while resolving the symbols. This feature is often used to modify behaviour of libc functions - fopen(), read(), write(), strcmp() etc, specially in case of obfuscated programs, where information is deliberately obfuscated to prevent static analysis, for instance, file path passed to fopen() or strings passed to strcmp().
+
+On Android, setting `LD_PRELOAD` is slightly different than compared to other Linux distributions: using the `setprop` command. Below you can see an example for an application with package name `com.foo.bar` (note the additional `wrap.` prefix):
```
TheDauntless: We can maybe add that this is because every application process is forked from zygote, so you would have to use LD_PRELOAD on the zygote process, which won't work of course. As a circumvention, android added the setprop functionality.

Document/0x05c-Reverse-Engineering-and-Tampering.md
```diff
@@ -1332,6 +1332,66 @@ The following approach can be used in order to patch the JavaScript file:
...
+
+###### LD_PRELOAD
+
+Above we looked into techniques which require some kind of modification of the application's code. A library can also be injected into a process using functionalities offered by the loader of the operating system. On Android, which is a Linux based OS, you can load an additional library by setting the `LD_PRELOAD` environment variable.
+
+The symbols loaded from the library passed using LD_PRELOAD always gets the precedence, i.e they are searched first by the loader while resolving the symbols. This feature is often used to modify behaviour of libc functions - fopen(), read(), write(), strcmp() etc, specially in case of obfuscated programs, where information is deliberately obfuscated to prevent static analysis, for instance, file path passed to fopen() or strings passed to strcmp().
```
cpholguera: I've restored the original suggestion, as this paragraph does not comply to the style guide.

Apart from that I'm not quite sure about last part.

> [...] information is deliberately obfuscated to prevent static analysis, for instance, file path passed to fopen()

I assume the use case here is: you might be able to patch your program (or not) but definitely you cannot _patch the original libc_. However, you need to get insights on its function parameters. Because `fopen` will open the file but you won't know which file. So you can override `fopen` including a `print`. 

Please take a look and let me know what you think. Gladly we can also discuss it over Slack :)

```suggestion
As the [ld.so man page](http://man7.org/linux/man-pages/man8/ld.so.8.html "LD.SO man page") states, symbols loaded from the library passed using `LD_PRELOAD` always get precedence, i.e they are searched first by the loader while resolving the symbols, effectively overriding the original ones. This feature is often used to inspect the input parameters of some commonly used libc functions such as `fopen`, `read`, `write`, `strcmp`, etc., specially in obfuscated programs, where understanding their behavior may be challenging. Therefore, having an insight on which files are being opened or which strings are being compared may be very valuable. The key idea here is "function wrapping", meaning that you cannot patch system calls e.g. libc's `fopen` but you can override (wrap) it including custom code that will print the input parameters for you but still call the original `fopen`.```

File: Document/0x05c-Reverse-Engineering-and-Tampering.md [hunk 4]
```diff
@@ -1332,6 +1332,66 @@ The following approach can be used in order to patch the JavaScript file:
...
+
+On Android, setting `LD_PRELOAD` is slightly different than compared to other Linux distributions: using the `setprop` command. Below you can see an example for an application with package name `com.foo.bar` (note the additional `wrap.` prefix):
+
+```bash
+$ setprop wrap.com.foo.bar LD_PRELOAD=/data/local/tmp/libpreload.so
+```
+
+> In the latest Android versions, you need to disable the SELinux to make LD_PRELOAD work, which may require root, as the library to be preloaded may not have SELinux context assigned.
```
cpholguera: Could you please include the API level? Something like ", starting on Android X (API level XX), you need to disable...

May require root? or does require root?

```suggestion
> Please note that if the library to be preloaded does not have SELinux context assigned, in the latest Android versions, you need to disable SELinux to make `LD_PRELOAD` work, which may require root.
```

su-vikas:
thanks for the suggestions, I have made fixes as suggested. And sorry it took too long to push updates :).

Issue #1662, opened 2020-02-25
Encrypted Shared Preferences not covered in MSTG-STORAGE-1 and MSTG-STORAGE-2


galapogos:
MSTG-STORAGE-1 and MSTG-STORAGE-2 lists proper usage of Shared Preferences, but doesn't provide guidelines for the new Encrypted Shared Preferences, which automatically encrypt the contents of the XML file.

cpholguera:
Hi @galapogos , thanks for pointing that out! If you have experience using that feature, would you mind opening a PR including what's missing? That'd be a great help for us, we'll gladly review it :)

galapogos:
I actually don't have experience using this, however I recently encountered an app using this while doing an assessment of it, and wanted to refer to MSTG on testing it, but didn't find anything, hence the issue. However, I did note that I could simply hook on the API call and get the decrypted data.

A-AFTAHI:
I think that testing encrypted that is similar either they're stored in Shared Preferences or elsewhere. don't you think? Perhaps adding a section specifically for Shared_Prefs would be Good.

cpholguera:
On a first step we should add a reference on usage of encrypted shared preferences in the current section. After that we could think if we need a separate section or a better sub-sections structure, I'm not sure if the current static and dynamic analysis sections even match.

cpholguera:
https://developer.android.com/topic/security/data

lwierzbicki:
@cpholguera I think I can work on it, as mentioned in issues #947  and #949.

cpholguera:
That'd be perfect, thank you so much!

jadeboer:
The mechanism being proposed by the new security lib is super simple to use, but the lib itself is an alpha. 

However, I would suggest the MSTG recommend its use when it is generally released. I would not recommend to any dev teams that they make use of alpha libraries in general.

https://developer.android.com/topic/security/data#edit-shared-preferences

lwierzbicki:
@jadeboer thanks for being thorough :) Lib will make things easier in the future. But I've seen implementations which use the same pattern - master key is stored in Keystore (or at the end it was advised to be stored in Keystore ;) ) and it is used to encrypt keys and values (which are encoded with base64 and saved in shared preferences - app container).

jadeboer:
Yep, the pattern is useful and this new security lib will also simplify consistent implementation of the pattern for developers. I just wanted to caution against recommending use of an alpha lib =)

[...]

PR #1664, opened 2020-03-05
(WIP) Update to Android cryptography

### LINKED ISSUE #947: Android Oreo & P updates: changes in cryptography
There have been nice updates for Android O & P in terms of how cryptography is handled: more hmacs and ciphers! Time to update the MSTG at the cryptography for Android.
 https://developer.android.com/about/versions/oreo/android-8.1 & https://developer.android.com/preview/features/security
https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html 
https://developer.android.com/training/articles/keystore

### LINKED ISSUE #949: Android P update: hardware security module support
With android P the hardware security module support has improved. Time to update data protection and cryptography chapter! https://developer.android.com/preview/features/security

lwierzbicki:
TODO list:
Closes #947 
- [x] update crypto/signing recommendations for android (what not to use and what to use)
- [x] update key derivation functions (recommend right ones, derecommend wrong/sweaker ones)
- [x] verify that provider selection is no longer recommended anywhere
- [x] keychain and keystore (key goal) updates (E..g what to use and what not to use)

Closes #949 
- [x] cover strongbox (And the requirement for a fallback for older devices
- [x] cover titan in terms of programmatic interaction if necessary.

Issue #1662 (only part job done)
- [x] encrypted shared preferences (how stored data in shared preferneces) 

MASVS issue 382
- [x] definition of hardcoded

I use comments in text like this```[//]: # (comment)```. This helps me add some notes to what needs to be done ;)

Notes:
* Provider selection is according to presented example (we have 1 example with BKS/BC and a few with Keystore)
* Definition of hardcoded was added to MSTG-CRYPTO-01 for Android.


lwierzbicki:
Added:

- Definition of hardcoded to 0x04g-Testing-Cryptography.
- Storing encrypted keys in SharedPreferences and encrypt/decrypt them by master key stored in Android Keystore - example.

Still in progress:

- review of keystore/keychain description
- strongbox, titan and hardware backed
- review of Android test cases.

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -273,9 +273,75 @@ You can use stored keys in one of two modes:
...
+Note that if you take a rooted device, or unpatched device, or a patched (e.g. repackaged) application into account as a threat to the data, it might be better to encrypt the salt with a key in the `AndroidKeystore`. Afterwards the Password-based Encryption (PBE) key is generated using the recommended `PBKDF2WithHmacSHA1` algorithm till Android 8.0 (API level 26). From there on, it is best to use `PBKDF2withHmacSHA256`, which will end up with a different key size.
```
sushi2k, line 300:
```suggestion
Note that if you take a rooted device or a patched (e.g. repackaged) application into account as a threat to the data, it might be better to encrypt the salt with a key that is placed in the `AndroidKeystore`. The Password-Based Encryption (PBE) key is generated using the recommended `PBKDF2WithHmacSHA1` algorithm, till Android 8.0 (API level 26). For higher API levels, it is best to use `PBKDF2withHmacSHA256`, which will end up with a longer hash value.
```

Should be the hash value instead of the key, right?

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -273,9 +273,75 @@ You can use stored keys in one of two modes:
...
+
+    //Generate the salt
+    byte[] salt = new byte[saltLength];
+    random.nextBytes(salt);
+
+    KeySpec keySpec = new PBEKeySpec(password.toCharArray(), salt, iterationCount, keyLength);
+    SecretKeyFactory keyFactory = SecretKeyFactory.getInstance("PBKDF2WithHmacSHA1");
+    byte[] keyBytes = keyFactory.generateSecret(keySpec).getEncoded();
+    return new SecretKeySpec(keyBytes, "AES");
+}
```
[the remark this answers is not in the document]
lwierzbicki: Code came from 0x05e-Testing-Cryptography - I moved part of a text to separate storing part from other crypto topics.  I need to revise chapter 0x05d (as I need want to put there hardware modules for storing key material).

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: Right, as you define key size as a function argument.

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
lwierzbicki: From my perspective, I would recommend removing that part of code:
* 1st we should discourage writing crypto by developers 
* 2nd we should recommend using built-in mechanism

lwierzbicki:
I pushed some updates - which should finalize this PR.
However, during working on that PR I discovered the following areas for further improvements:
* Dynamic Analysis for MSTG-CRYPTO-01
* Dynamic Analysis for MSTG-CRYPTO-2, MSTG-CRYPTO-3 and MSTG-CRYPTO-4
* Dynamic Analysis for MSTG-CRYPTO-05
I would recommended raising an issues (I can do that) for them and taking them separately (as this PR is growing).

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
commjoen: Why would it be wrong to write this? It's still not roling your own and it is perfectly supported by Android.If you need a password based key derivation function, then it should be fine. I rather use `PBKDF2withHmacSHA256` as is available according to https://developer.android.com/reference/javax/crypto/SecretKeyFactory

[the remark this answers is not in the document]
lwierzbicki: That's a good point. I need to think where to put it then.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,70 @@
...
+
+- generating of a key
+- using of a key
+- storing of a key
+- archiving of a key
+- deleting of a key
+
+> Please note that storing of a key is analyzed in the chapter "[Testing Data Storage](0x05d-Testing-Data-Storage.md)".
+
+These phases are managed by Keystore/KeyChain system. However how the system works depends on how application developer used it. Then, you should focus on functions which are used by application developer. You should identify and verify following functions:
```
commjoen, line 25: let's add Key-rotation here as well: after all every key has a given lifetime given the security put into securing it and the need for attackers to grab it I(e.g. risk of the data being leaked)
.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,70 @@
...
+
+- [Key generation](0x05e-Testing-Cryptography.md#key-generation)
+- [Random number generation](0x05e-Testing-Cryptography.md#random-number-generation)
+
+You can go further and verify how good is a key by performing [key attestation](0x05e-Testing-Cryptography.md#key-attestation).
```
commjoen, line 30: i thought key attestation was mostly about verifying the integrity of the Secure world in android. Not necessarily whether the key itself was ok? And is it already broadly available?

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,70 @@
...
+
+Apps that target modern API levels, went through the following changes:
+
+- For Android 7.0 (API level 24) and above [the Android Developer blog shows that](https://android-developers.googleblog.com/2016/06/security-crypto-provider-deprecated-in.html "Security provider Crypto deprecated in Andorid N"):
+  - It is recommended to stop specifying a security provider. Instead, always use a patched security provider.
+  - The support for the `Crypto` provider has dropped and the provider is deprecated. The same applies to its `SHA1PRNG` for secure random.
+- For Android 8.1 (API level 27) and above the [Developer Documentation](https://developer.android.com/about/versions/oreo/android-8.1 "Cryptography updates") shows that:
+  - Conscrypt, known as `AndroidOpenSSL`, is preferred above using Bouncy Castle and it has new implementations: `AlgorithmParameters:GCM` , `KeyGenerator:AES`, `KeyGenerator:DESEDE`, `KeyGenerator:HMACMD5`, `KeyGenerator:HMACSHA1`, `KeyGenerator:HMACSHA224`, `KeyGenerator:HMACSHA256`, `KeyGenerator:HMACSHA384`, `KeyGenerator:HMACSHA512`, `SecretKeyFactory:DESEDE`, and `Signature:NONEWITHECDSA`.
```
commjoen, line 38: please note that this works well/best, if you patch your security provider. I would recommend to link to that as well.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,70 @@
...
+  - You should not use the `IvParameterSpec.class` anymore for GCM, but use the `GCMParameterSpec.class` instead.
```
commjoen, line 39: IT would be good to detail that a bit, as there are a few problems with GCM (maybe we have it elsewhere with the guide):
the Nonce should NOT be repeated for the same key (as it is no IV), so there is a statemachine involved and/lor you need to change the key every time as well (e.g. use envelope-encryption). and there are limits to ohw many times you can use it.

Document/0x05e-Testing-Cryptography.md
```diff
@@ -39,25 +93,21 @@ provider: HarmonyJSSE1.0 (Harmony JSSE Provider)
...
+For some applications that support older versions of Android (e.g.: only used versions lower than Android 7.0 (API level 24)), bundling an up-to-date library may be the only option. Spongy Castle (a repackaged version of Bouncy Castle) is a common choice in these situations. Repackaging is necessary because Bouncy Castle is included in the Android SDK. The latest version of [Spongy Castle](https://rtyley.github.io/spongycastle/ "Spongy Castle") likely fixes issues encountered in the earlier versions of [Bouncy Castle](https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html "CVE Details Bouncy Castle") that were included in Android. Note that the Bouncy Castle libraries packed with Android are often not as complete as their counterparts from the [legion of the Bouncy Castle](https://www.bouncycastle.org/java.html "Bouncy Castle in Java"). Lastly: bear in mind that packing large libraries such as Spongy Castle will often lead to a multidexed Android application.
+
```
commjoen, line 109: is Libsodium a valid alternative here to get less size in the app?

Document/0x05e-Testing-Cryptography.md
```diff
@@ -152,147 +202,13 @@ This sample creates the RSA key pair with a key size of 4096-bit (i.e. modulus s
...
-
-When keys are generated and used within the `AndroidKeyStore` and the `KeyInfo.isinsideSecureHardware` returns `true`, then we know that we cannot just dump the keys nor monitor its cryptographic operations. It becomes debatable what will be eventually more safe: using `PBKDF2withHmacSHA256` to generate a key that is still in reachable dumpable memory, or using the `AndroidKeyStore` for which the keys might never get into memory. With Android 9 (API level 28) we see that additional security enhancements have been implemented in order to separate the TEE from the `AndroidKeyStore` which make it favorable over using `PBKDF2withHmacSHA256`. However, more testing and investigating will take place on that subject in the near future.
```
commjoen, line 265: the above paragraph still is pretty strong to me, why would we remove it?

Document/0x05e-Testing-Cryptography.md
```diff
@@ -349,36 +265,31 @@ For the security analysis perspective the analysts may perform the following che
...
-Next, when you have a predictable key derivation function based on identifiers which are accessible to other applications, the attacker only needs to find the KDF and apply it to the device in order to find the key. Lastly, storing encryption keys publicly also is highly discouraged as other applications can have permission to read the public partition and steal the keys.
+Verify that symmetric keys in all identified instances are not hardcoded. Check if symmetric keys are not:
 
-#### Static Analysis
+- part of application resources
+- part of a static identifier (which might not be clearly within the definition)
+- hardcoded in code
```
commjoen, line 291: Maybe add something like "part of a derivation function that always ends up with the same result"?

commjoen:
Your review and our conversation at Slack spiked the following items that woudl need a ticket as well:
- GCMParameterSpec is only addressed a bit in anti-reverse engineering, while this needs more love in crypto
- Signing needs more love (RSA-PSS as prefered over the older PKCS1.5 padding)
- padding of RSA needs more love in general 
- curve usage needs more love

Document/0x05e-Testing-Cryptography.md
lwierzbicki: I devided it. I used the part regarding AndroidKeyStore and SharedPreferences. But need to put back part about `PBKDF2withHmacSHA256`. Currently it should be in 0x05d-Testing-Data-Storage.md#Storing-a-Key-example.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: My bad, I think about something else here (key lifecycle) and link it incorrectly.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: In recommendation section (Document/0x05e-Testing-Cryptography.md#recommendation) I added that security provided should be updated (and provide a link to Android docs). Do you think to add it here also?

commjoen: I guess so :), just a quick reference to the recommendation and the example there.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Can be, but I need to check that first.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: I started to consider different angles here. Maybe, I overthink this, but would be better to say that 'value which can be derived from known values' This should cover "part of a derivation function that always ends up with the same result" (which can be a narrow case). 
Example:
* function(application_name)
* function(system_variable)
* function which multiply by 0

What do you think?

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Good point. That topic need `more love` also (at least one section level 4).

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Done.

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
lwierzbicki: Reintroduces in key generation section.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Added to list of functionalities and added to #1689

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Added to #1689

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Additionally, I moved key attestation to Testing Data Storage as it is related closely to hardware-backed storage of keys.

Document/0x05e-Testing-Cryptography.md
commjoen: I guess we mostly need to check for an alternative when the NDK is used instead of the SDK. Not sure what fits there currently as I haven't visited the NDK for a long time.

Document/0x05e-Testing-Cryptography.md
commjoen: yes, sounds legit.

commjoen:
> Added:
> 
> * Definition of hardcoded to 0x04g-Testing-Cryptography.
> * Storing encrypted keys in SharedPreferences and encrypt/decrypt them by master key stored in Android Keystore - example.
> 
> Still in progress:
> 
> * review of keystore/keychain description
> * strongbox, titan and hardware backed
> * review of Android test cases.

BTW: stronger would still be to have the master key on the server after some sort of check-in. So that you can monitor account usage at least at the server side.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Added to #1689 - as it requires further investigation.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Corrected.

lwierzbicki:
Corrected hardcoded definition. 
`BTW: stronger would still be to have the master key on the server after some sort of check-in. So that you can monitor account usage at least at the server side.` - I provided a list in section `Storing a Key - example`  of 0x05d-Testing-Data-Storage.

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -271,11 +271,104 @@ You can use stored keys in one of two modes:
...
+
+Although the key attestation process can be implemented within the application directly but it is recommended that it should be implemented at the server-side for security reasons. The following are the high-level guidelines for the secure implementation of Key Attestation:
+
+- The server should initiate the key attestation process by creating a random number securely using CSPRNG(Cryptographically Secure Random Number Generator) and the same should be sent to the user as a challenge.
+- The client should call the `setAttestationChallenge` API with the challenge received from the server and should then retrieve the attestation certificate chain using the `KeyStore.getCertificateChain` method.
+- The attestation response should be sent to the server for the verification and following checks should be performed for the verification of the key attestation response:
+  - Verify the certificate chain, up to the root and perform certificate sanity checks such as validity, integrity and trustworthiness.
```
commjoen, line 297: Don't forget to check against https://android.googleapis.com/attestation/status to see if one of the certificates of the chain is part of the CRL.

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -271,11 +271,104 @@ You can use stored keys in one of two modes:
...
+  - Check if the root certificate is signed with the Google attestation root key which makes the attestation process trustworthy.
+  - Extract the attestation certificate extension data, which appears within the first element of the certificate chain and perform the following checks:
+    - Verify that the attestation challenge is having the same value which was generated at the server while initiating the attestation process.
+    - Verify the signature in the key attestation response.
+    - Now check the security level of the Keymaster to determine if the device has secure key storage mechanism. Keymaster is a piece of software that runs in the security context and provides all the secure keystore operations. The security level will be one of `Software`, `TrustedEnvironment` or `StrongBox`.
```
commjoen, line 302: note that software version can be done by any... so be careful then. take TE and SB as a real security hook, or you will need to re-evaluate the threatmodel in place as it is one big compromisable software party.

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -271,11 +271,104 @@ You can use stored keys in one of two modes:
...
+
+###### Key Attestation
+
+For the applications which heavily rely on Android Keystore for business-critical operations such as multi-factor authentication through cryptographic primitives, secure storage of sensitive data at the client-side, etc. Android provides the feature of [Key Attestation](https://developer.android.com/training/articles/security-key-attestation "Key Attestation") which helps to analyze the security of cryptographic material managed through Android Keystore. From Android 8.0 (API level 26), the key attestation was made mandatory for all new(Android 7.0 or higher) devices that need to have device certification for Google suite of apps, such devices use attestation keys signed by the [Google hardware attestation root certificate](https://developer.android.com/training/articles/security-key-attestation#root_certificate "Google Hardware Attestation Root Certificate") and the same can be verified while key attestation process.
```
commjoen, line 288: maybe good to go over the rest and see how you can use it for checkup of the security of the device> and then move that for a part into tamper detection and such...

cpholguera:
Hi @lwierzbicki, if this PR should close the above mentioned issues please substitute the word "Issue" with "closes" in your PR description so that they are automatically linked and we can see the dependencies (and here in the PR in the section "Linked Issues"). Thanks!

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: Added.

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: Not sure what you had in mind here. In Key Attestation section there is no `software version` keyword and I'm not able to find relation.

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: That leads me to consideration of some kind Android device security checklist. I think it would be good to take that  as a separate issue and define goals and objectives which we want to achieve. What do you think?

Document/0x05d-Testing-Data-Storage.md
commjoen: ah no, i mean security level being "software". in that case it might be faked by the device and/or an attacker. there were some notes / warnings 👍 in https://developer.android.com/training/articles/security-key-attestation as well:

```Most likely, the device launched with an Android version less than 7.0 and it does not support hardware attestation. In this case, Android has a software implementation of attestation which produces the same sort of attestation certificate, but signed with a key hardcoded in Android source code. Because this signing key is not a secret, the attestation could have been created by an attacker pretending to provide secure hardware.```

File: Document/0x05d-Testing-Data-Storage.md (hunk 12 above)
Comment (commjoen) [reply Document/0x05d-Testing-Data-Storage.md:288]: I believe one should not be separated from another: key attestation only makes sense if you evaluate ```RootOfTrust ::= SEQUENCE {
    verifiedBootKey  OCTET_STRING,
    deviceLocked  BOOLEAN,
    verifiedBootState  VerifiedBootState,
    verifiedBootHash OCTET_STRING,
}

VerifiedBootState ::= ENUMERATED {
    Verified  (0),
    SelfSigned  (1),
    Unverified  (2),
    Failed  (3),
}```
because that will tell whether the device has been played with. This can then tell something about  what level of trust you can put into the device.

File: Document/0x05d-Testing-Data-Storage.md (hunk 11 above)
Comment (lwierzbicki) [reply Document/0x05d-Testing-Data-Storage.md]: That is clear :) I tried clarifying this - device which support hardware-key attestation and what to check.

File: Document/0x05d-Testing-Data-Storage.md (hunk 12 above)
Comment (lwierzbicki) [reply Document/0x05d-Testing-Data-Storage.md:288]: That is clear. I added additional check which should verify device's status and simplify one of existing.

File: Document/0x05d-Testing-Data-Storage.md [hunk 13]
```diff
@@ -271,11 +271,107 @@ You can use stored keys in one of two modes:
...
+
+- The `algorithm` parameter Specifies the cryptographic algorithm with which the key is used
+- The `keySize` parameter Specifies the size, in bits, of the key, measuring in the normal way for the key's algorithm
+- The `digest` parameter Specifies the digest algorithms that may be used with the key to perform signing and verification operations
+
+###### Older KeyStore Implementations
 
 Older Android versions don't include KeyStore, but they *do* include the KeyStore interface from JCA (Java Cryptography Architecture). You can use KeyStores that implement this interface to ensure the secrecy and integrity of keys stored with KeyStore; BouncyCastle KeyStore (BKS) is recommended. All implementations are based on the fact that files are stored on the filesystem; all files are password-protected.
 To create one, you can use the `KeyStore.getInstance("BKS", "BC") method`, where "BKS" is the KeyStore name (BouncyCastle Keystore) and "BC" is the provider (BouncyCastle). You can also use SpongyCastle as a wrapper and initialize the KeyStore as follows: `KeyStore.getInstance("BKS", "SC")`.
```
commjoen, line 377: BCFKS would be more secure here, but requires a later version of java, would that work with spongy castle? https://cryptosense.com/blog/bouncycastle-keystore-security/

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+- KeyChain - see the section [KeyChain](0x05d-Testing-Data-Storage.md#keychain) in the chapter "Testing Data Storage"
 
 Android cryptography APIs are based on the Java Cryptography Architecture (JCA). JCA separates the interfaces and implementation, making it possible to include several [security providers](https://developer.android.com/reference/java/security/Provider.html "Android Security Providers") that can implement sets of cryptographic algorithms. Most of the JCA interfaces and classes are defined in the `java.security.*` and `javax.crypto.*` packages. In addition, there are Android specific packages `android.security.*` and `android.security.keystore.*`.
 
+KeyStore and KeyChain provides APIs for storing and using keys (behind the scene, KeyChain API uses KeyStore system). These systems allow to administer the full lifecycle of the cryptographic keys. Requirements and guidance for implementation of cryptographic key management can be found in [Key Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Key_Management_Cheat_Sheet.html "Key Management Cheat Sheet"). We can identify following phases:
```
commjoen, line 15:
```suggestion
KeyStore and KeyChain provide APIs for storing and using keys (behind the scene, KeyChain API uses KeyStore system). These systems allow to administer the full lifecycle of the cryptographic keys. Requirements and guidance for implementation of cryptographic key management can be found in [Key Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Key_Management_Cheat_Sheet.html "Key Management Cheat Sheet"). We can identify following phases:
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+  - It is recommended to stop specifying a security provider. Instead, always use a [patched security provider](0x05e-Testing-Cryptography.md#updating-provider).
+  - The support for the `Crypto` provider has dropped and the provider is deprecated. The same applies to its `SHA1PRNG` for secure random.
+- For Android 8.1 (API level 27) and above the [Developer Documentation](https://developer.android.com/about/versions/oreo/android-8.1 "Cryptography updates") shows that:
+  - Conscrypt, known as `AndroidOpenSSL`, is preferred above using Bouncy Castle and it has new implementations: `AlgorithmParameters:GCM` , `KeyGenerator:AES`, `KeyGenerator:DESEDE`, `KeyGenerator:HMACMD5`, `KeyGenerator:HMACSHA1`, `KeyGenerator:HMACSHA224`, `KeyGenerator:HMACSHA256`, `KeyGenerator:HMACSHA384`, `KeyGenerator:HMACSHA512`, `SecretKeyFactory:DESEDE`, and `Signature:NONEWITHECDSA`.
+  - You should not use the `IvParameterSpec.class` anymore for GCM, but use the `GCMParameterSpec.class` instead.
+  - Sockets have changed from `OpenSSLSocketImpl` to `ConscryptFileDescriptorSocket`, and `ConscryptEngineSocket`.
+  - `SSLSession` with null parameters give an NullPointerException.
```
commjoen, line 40:
```suggestion
  - `SSLSession` with null parameters give a `NullPointerException`.
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+  - You need to have large enough arrays as input bytes for generating a key otherwise, an InvalidKeySpecException is thrown.
```
commjoen, line 41:
```suggestion
  - You need to have large enough arrays as input bytes for generating a key otherwise, an `InvalidKeySpecException` is thrown.
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+  - If a Socket read is interrupted, you get an `SocketException`.
```
commjoen, line 42:
```suggestion
  - If a Socket read is interrupted, you get a `SocketException`.
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+- For Android 9 (API level 28) and above the [Android Developer Blog](https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html "Cryptography Changes in Android P") shows even more aggressive changes:
```
commjoen, line 43:
```suggestion
- For Android 9 (API level 28) and above the [Android Developer Blog](https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html "Cryptography Changes in Android P") shows even more changes:
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+  - You get a warning if you still specify a provider using the `getInstance` method and you target any API below 28. If you target Android 9 (API level 28) or above, you get an error.
+  - The `Crypto` provider is now removed. Calling it will result in a `NoSuchProviderException`.
+- For Android 10 (API level 29) the [Developer Documentation](https://developer.android.com/about/versions/10/behavior-changes-all#security "Security Changes in Android 10") list all network security changes.
+
+#### Recommendation
+
+The following list of recommendations should be considered during app examination:
+
+- You should ensure that the best practices outlined in the "[Cryptography for Mobile Apps](0x04g-Testing-Cryptography.md)" chapter are followed.
+- You should ensure that provider has the latest updates - [Updating security provider](https://developer.android.com/training/articles/security-gms-provider "Updating security provider").
```
commjoen, line 53:
```suggestion
- You should ensure that the security provider has the latest updates - [Updating security provider](https://developer.android.com/training/articles/security-gms-provider "Updating security provider").
```

commjoen: maybe call it security provider everywhere?

Document/0x05e-Testing-Cryptography.md
```diff
@@ -150,264 +205,211 @@ KeyPair keyPair = keyPairGenerator.generateKeyPair();
...
-
-#### Static Analysis
+A symmetric encryption key can be generated from the passphrase by using the Password Based Key Derivation Function version 2 (PBKDF2). This cryptographic protocol is designed to generate secure and non brute-forceable keys. The code listing below illustrates how to generate a strong encryption key based on a password.
 
-Locate uses of the cryptographic primitives in code. Some of the most frequently used classes and interfaces:
+```java
+public static SecretKey generateStrongAESKey(char[] password, int keyLength)
+{
+    //Initiliaze objects and variables for later use
+    int iterationCount = 10000;
```
commjoen, line 214: maybe make this larger and explain what to do depending on the device profile?

Document/0x05e-Testing-Cryptography.md
```diff
@@ -150,264 +205,211 @@ KeyPair keyPair = keyPairGenerator.generateKeyPair();
...
+This test case focuses on random values used by application. Following checks should be performed:
 
-The last three options are to use hardcoded encryption keys in the source code, having a predictable key derivation function based on stable attributes, and storing generated keys in public places like `/sdcard/`. Obviously, hardcoded encryption keys are not the way to go. This means every instance of the application uses the same encryption key. An attacker needs only to do the work once, to extract the key from the source code - whether stored natively or in Java/Kotlin. Consequently, an attacker can decrypt any other data which was encrypted by the application.
-Next, when you have a predictable key derivation function based on identifiers which are accessible to other applications, the attacker only needs to find the KDF and apply it to the device in order to find the key. Lastly, storing encryption keys publicly also is highly discouraged as other applications can have permission to read the public partition and steal the keys.
+- identify all instances where random values are used and all instances of random number generators
```
commjoen, line 361:
```suggestion
- identify all instances where random values are used and all instances of random number generators are of `Securerandom`
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -291,6 +387,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
 - Use the `store(OutputStream stream, char[] password)` function to store the KeyStore to disk with a password. Make sure that the password is provided by the user, not hard-coded.
 
+##### Storing a Key - example
+
+To mitigate unauthorized use of keys on the Android device, Android KeyStore lets apps specify authorized uses of their keys when generating or importing the keys. Once made, authorizations cannot be changed.
+
+Storing a Key - from most secure to least secure:
+
+- all keys are stored on server and available authentication
+- master key is stored on server and use to encrypt other keys, which are stored in Android KeyStore/SharedPreferences
```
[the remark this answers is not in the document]
lwierzbicki: I should look closer when I use copy+paste ;)

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -291,6 +387,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+- the key is derived each time from user provided passphrase
```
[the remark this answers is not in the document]
lwierzbicki: Corrected

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -291,6 +387,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+- the key is stored in hardware-backed Android KeyStore
+- the key is stored in software implementation of Android KeyStore
+- master key is stored in Android Keystore and use to encrypt other keys, which are stored in SharedPreferences
+- all keys are stored in SharedPreferences
+- [not recommended] hardcoded encryption keys in the source code
+- [not recommended] predictable key derivation function based on stable attributes
+- [not recommended] stored generated keys in public places (like `/sdcard/`)
```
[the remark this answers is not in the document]
lwierzbicki: I think inspecting patch level suits better for testing MASVS 7.5 (where we check for know vulnerabilities). In such case, example of Storing a Key could expand too much. What do you think?

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Changed to security provider.

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -291,6 +387,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+
+The most secure way of handling key material, is simply never storing it on the device. This means that the user should be prompted to input a passphrase every time the application needs to perform a cryptographic operation. Although this is not the ideal implementation from a user experience point of view and construction of passphrase (prone to human error), it is however the most secure way of handling key material. The reason is because key material will only be available in an array in memory while it is being used. Once the key is not needed anymore, the array can be zeroed out. This minimizes the attack window as good as possible. No key material touches the filesystem and no passphrase is stored. However, note that some ciphers do not properly clean up their byte-arrays. For instance, the AES Cipher in BouncyCastle does not always clean up its latest working key. Next, BigInteger based keys (e.g. private keys) cannot be removed from the heap nor zeroed out just like that. Last, take care when trying to zero out the key.
```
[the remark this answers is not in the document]
lwierzbicki: Can you elaborate more about this case? I'm not sure if I understand correctly. 
Sections which (based on my understanding) suit here:
* references to work about prompting a passphrase every time
* describing more about zeroing key material from memory
As this was previously in MSTG (I just reorganize some content), then maybe I don't see the full picture here.

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
commjoen: Sounds indeed like a good plan. can you perhaps alter the todo-issue about this please? Then we can resolve this comment :).

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
commjoen: Both need examples of how to do it properly and pitfalls what to look for when stuff goes wrong, think off:
- passphrases with salting (salt wrongly saved, passphrase too short)
- having copies of the byte-array hanging arround, etc.

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
lwierzbicki: Created a new issue for MSTG-CODE-5: #1713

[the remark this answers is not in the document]
commjoen: Thanks!

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: As discussed, I updated #1689.

Document/0x05e-Testing-Cryptography.md
lwierzbicki: Added references to NIST SP (mainly in 0x04g) and some minor fixes.

Document/0x05d-Testing-Data-Storage.md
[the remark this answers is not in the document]
lwierzbicki: Extended that section a bit (by putting references) and reorganized as discussed.

Document/0x04g-Testing-Cryptography.md
```diff
@@ -93,7 +100,7 @@ Cryptographic algorithms (such as symmetric encryption or some MACs) expect a se
...
+Ensure that passwords aren't directly passed into an encryption function. Instead, the user-supplied password should be passed into a KDF to create a cryptographic key. Choose an appropriate iteration count when using password derivation functions. For example, [NIST recommends an iteration count of at least 10,000 for PBKDF2](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5 "NIST Special Publication 800-63B") and [for critical keys where user-perceived performance is not ciritical - 10,000,000](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf "NIST Special Publication 800-132"). For critical keys, it is recommended to consider implementation of algorithms recognized by [Password Hashing Competition (PHC)](https://password-hashing.net/ "PHC") like [Argon2](https://github.com/p-h-c/phc-winner-argon2 "Argon2").
```
sushi2k, line 103:
```suggestion
Ensure that passwords aren't directly passed into an encryption function. Instead, the user-supplied password should be passed into a KDF to create a cryptographic key. Choose an appropriate iteration count when using password derivation functions. For example, [NIST recommends an iteration count of at least 10,000 for PBKDF2](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5 "NIST Special Publication 800-63B") and [for critical keys where user-perceived performance is not critical at least 10,000,000](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf "NIST Special Publication 800-132"). For critical keys, it is recommended to consider implementation of algorithms recognized by [Password Hashing Competition (PHC)](https://password-hashing.net/ "PHC") like [Argon2](https://github.com/p-h-c/phc-winner-argon2 "Argon2").
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -290,11 +290,107 @@ You can use stored keys in one of two modes:
...
...d. Keymaster Hardware Abstraction Layer (HAL) was introduced with device running Android 6 (API level 23). Application can verify if key is stored inside security hardware (by checking if `KeyInfo.isinsideSecureHardware` returns `true`). Devices running Android 9 (API level 28) and higher can have a `StrongBox Keymaster` module, an implementation of the Keymaster HAL that resides in a hardware security module which has its own CPU, Secure storage, a true random number generator and a mechanism to resist package tampering. To use this feature, `true` must be passed to the `setIsStrongBoxBacked` method in either the `KeyGenParameterSpec.Builder` class or the `KeyProtection.Builder` class when generating or importing keys using `AndroidKeystore`. To make sure that StrongBox is used during runtime check that `isInsideSecureHardware` returns `true` and that the system does not throw `StrongBoxUnavailableException` which get thrown if the StrongBox Keymaster isn't available for the given algorithm and key size associated with a key. Description of features on hardware-based keystore can be found on [AOSP pages](https://source.android.com/security/keystore "AOSP Hardware-based KeyStore").
```
sushi2k, line 301:
```suggestion
As mentioned before, hardware-backed Android KeyStore gives another layer to defense-in-depth security concept for Android. Keymaster Hardware Abstraction Layer (HAL) was introduced with Android 6 (API level 23). Applications can verify if the key is stored inside the security hardware (by checking if `KeyInfo.isinsideSecureHardware` returns `true`). Devices running Android 9 (API level 28) and higher can have a `StrongBox Keymaster` module, an implementation of the Keymaster HAL that resides in a hardware security module which has its own CPU, Secure storage, a true random number generator and a mechanism to resist package tampering. To use this feature, `true` must be passed to the `setIsStrongBoxBacked` method in either the `KeyGenParameterSpec.Builder` class or the `KeyProtection.Builder` class when generating or importing keys using `AndroidKeystore`. To make sure that StrongBox is used during runtime, check that `isInsideSecureHardware` returns `true` and that the system does not throw `StrongBoxUnavailableException` which gets thrown if the StrongBox Keymaster isn't available for the given algorithm and key size associated with a key. Description of features on hardware-based keystore can be found on [AOSP pages](https://source.android.com/security/keystore "AOSP Hardware-based KeyStore").
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -290,11 +290,107 @@ You can use stored keys in one of two modes:
...
+
+Keymaster HAL is an interface to hardware-backed component - Trusted Execution Environment (TEE) or a Secure Element (SE), which is used by Android Keystore. Example of such hardware-backed component is [Titan M](https://android-developers.googleblog.com/2018/10/building-titan-better-security-through.html "Building a Titan: Better security through a tiny chip").
```
sushi2k, line 303:
```suggestion
Keymaster HAL is an interface to hardware-backed components - Trusted Execution Environment (TEE) or a Secure Element (SE), which is used by Android Keystore. An example of such a hardware-backed component is [Titan M](https://android-developers.googleblog.com/2018/10/building-titan-better-security-through.html "Building a Titan: Better security through a tiny chip").
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -290,11 +290,107 @@ You can use stored keys in one of two modes:
...
+
+###### Key Attestation
+
+For the applications which heavily rely on Android Keystore for business-critical operations such as multi-factor authentication through cryptographic primitives, secure storage of sensitive data at the client-side, etc. Android provides the feature of [Key Attestation](https://developer.android.com/training/articles/security-key-attestation "Key Attestation") which helps to analyze the security of cryptographic material managed through Android Keystore. From Android 8.0 (API level 26), the key attestation was made mandatory for all new(Android 7.0 or higher) devices that need to have device certification for Google suite of apps, such devices use attestation keys signed by the [Google hardware attestation root certificate](https://developer.android.com/training/articles/security-key-attestation#root_certificate "Google Hardware Attestation Root Certificate") and the same can be verified while key attestation process.
```
sushi2k, line 307:
```suggestion
For the applications which heavily rely on Android Keystore for business-critical operations such as multi-factor authentication through cryptographic primitives, secure storage of sensitive data at the client-side, etc. Android provides the feature of [Key Attestation](https://developer.android.com/training/articles/security-key-attestation "Key Attestation") which helps to analyze the security of cryptographic material managed through Android Keystore. From Android 8.0 (API level 26), the key attestation was made mandatory for all new (Android 7.0 or higher) devices that need to have device certification for Google apps. Such devices use attestation keys signed by the [Google hardware attestation root certificate](https://developer.android.com/training/articles/security-key-attestation#root_certificate "Google Hardware Attestation Root Certificate") and the same can be verified through the key attestation process.
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -290,11 +290,107 @@ You can use stored keys in one of two modes:
...
+
+During key attestation, we can specify the alias of a key pair and in return, get a certificate chain, which we can use to verify the properties of that key pair. If the root certificate of the chain is [Google Hardware Attestation Root certificate](https://developer.android.com/training/articles/security-key-attestation#root_certificate "Google Hardware Attestation Root certificate") and the checks related to key pair storage in hardware are made it gives an assurance that the device supports hardware-level key attestation and the key is in hardware-backed keystore that Google believes to be secure. Alternatively, if the attestation chain has any other root certificate, then Google does not make any claims about the security of the hardware.
```
sushi2k, line 309:
```suggestion
During key attestation, we can specify the alias of a key pair and in return, get a certificate chain, which we can use to verify the properties of that key pair. If the root certificate of the chain is the [Google Hardware Attestation Root certificate](https://developer.android.com/training/articles/security-key-attestation#root_certificate "Google Hardware Attestation Root certificate") and the checks related to key pair storage in hardware are made it gives an assurance that the device supports hardware-level key attestation and the key is in the hardware-backed keystore that Google believes to be secure. Alternatively, if the attestation chain has any other root certificate, then Google does not make any claims about the security of the hardware.
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -290,11 +290,107 @@ You can use stored keys in one of two modes:
...
+
+- The server should initiate the key attestation process by creating a random number securely using CSPRNG(Cryptographically Secure Random Number Generator) and the same should be sent to the user as a challenge.
+- The client should call the `setAttestationChallenge` API with the challenge received from the server and should then retrieve the attestation certificate chain using the `KeyStore.getCertificateChain` method.
+- The attestation response should be sent to the server for the verification and following checks should be performed for the verification of the key attestation response:
+  - Verify the certificate chain, up to the root and perform certificate sanity checks such as validity, integrity and trustworthiness. Check on Google mantained [Certificate Revocation Status List](https://developer.android.com/training/articles/security-key-attestation#root_certificat "Certificate Revocation Status List") if none of the cerificates in the chain was revoked.
```
sushi2k, line 316:
```suggestion
  - Verify the certificate chain, up to the root and perform certificate sanity checks such as validity, integrity and trustworthiness. Check the [Certificate Revocation Status List](https://developer.android.com/training/articles/security-key-attestation#root_certificat "Certificate Revocation Status List") maintained by Google, if none of the certificates in the chain was revoked.
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -290,11 +290,107 @@ You can use stored keys in one of two modes:
...
+    - Verify that the attestation challenge is having the same value which was generated at the server while initiating the attestation process.
+    - Verify the signature in the key attestation response.
+    - Verify the security level of the Keymaster to determine if the device has secure key storage mechanism. Keymaster is a piece of software that runs in the security context and provides all the secure keystore operations. The security level will be one of `Software`, `TrustedEnvironment` or `StrongBox`. The client supports hardware-level key attestation if security level is `TrustedEnvironment` or `StrongBox` and attestation certificate chain contains a root certificate singed with Google attestation root key.
+    - Verify client's status to ensure full chain of trust - verified boot key, locked bootloader and verified boot state.
+    - Additionally, you can verify the key pair's attributes such as purpose, access time, authentication requirement, etc.
+
+> Note, if for any reason that process fails, it means that the key is not in security hardware. That does not mean that the key is compromised
```
sushi2k, line 325:
```suggestion
> Note, if for any reason that process fails, it means that the key is not in security hardware. That does not mean that the key is compromised.
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -310,6 +406,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+
+Storing a Key - from most secure to least secure:
+
+- the key is stored in hardware-backed Android KeyStore
+- all keys are stored on server and are available after strong authentication
+- master key is stored on server and use to encrypt other keys, which are stored in Android SharedPreferences
+- the key is derived each time from a strong user provided passphrase with sufficient length and salt
+- the key is stored in software implementation of Android KeyStore
+- master key is stored in Android Keystore and use to encrypt other keys, which are stored in SharedPreferences
+- all keys are stored in SharedPreferences
```
sushi2k, line 421: this should be "not recommended"

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -310,6 +406,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+- [not recommended] hardcoded encryption keys in the source code
+- [not recommended] predictable key derivation function based on stable attributes
+- [not recommended] stored generated keys in public places (like `/sdcard/`)
+
+The most secure way of handling key material, is simply never storing it on the device. That can be achieved by using [hardware-backed Android KeyStore](0x05d-Testing-Data-Storage.md#hardware-backed-android-keystore) if device is running Android 7.0 (API level 24) and above with available hardware component (Trusted Execution Environment (TEE) or a Secure Element (SE)). That can be check by using guidelines provided for [the secure implementation of Key Attestation](0x05d-Testing-Data-Storage.md#key-attestation). If hardware component is not available and/or support for Android 6.0 (API level 23) and below is required, then that can be achieved by storing a key on remote server and make a key available after authentication. Limitiation is that app requires always online work.
```
sushi2k, line 426: The consequences of hosting the keys on the server should be more emphasized to make it clear, as this is a severe downside in some app architectures:

```suggestion
The most secure way of handling key material, is simply never storing it on the device. That can be achieved by using [hardware-backed Android KeyStore](0x05d-Testing-Data-Storage.md#hardware-backed-android-keystore) if device is running Android 7.0 (API level 24) and above with available hardware component (Trusted Execution Environment (TEE) or a Secure Element (SE)). That can be check by using guidelines provided for [the secure implementation of Key Attestation](0x05d-Testing-Data-Storage.md#key-attestation). If hardware component is not available and/or support for Android 6.0 (API level 23) and below is required, then that can be achieved by storing a key on remote server and make a key available after authentication. 

> Please note that if the keys are stored on the server, the app need to be online to decrypt the data. This might be a limitation in some use case of mobile apps and should be carefully thought through as this becomes part of the architecture of the app.
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -310,6 +406,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+To mitigate unauthorized use of keys on the Android device, Android KeyStore lets apps specify authorized uses of their keys when generating or importing the keys. Once made, authorizations cannot be changed.
+
+Storing a Key - from most secure to least secure:
+
+- the key is stored in hardware-backed Android KeyStore
+- all keys are stored on server and are available after strong authentication
+- master key is stored on server and use to encrypt other keys, which are stored in Android SharedPreferences
+- the key is derived each time from a strong user provided passphrase with sufficient length and salt
+- the key is stored in software implementation of Android KeyStore
+- master key is stored in Android Keystore and use to encrypt other keys, which are stored in SharedPreferences
```
sushi2k, line 420:
```suggestion
- master key is stored in software implementation of Android Keystore and use to encrypt other keys, which are stored in SharedPreferences
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -310,6 +406,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
...int of view and construction of passphrase, which is prone to human error. However both approaches make a key available in an array in memory while it is being used and when the key is not needed anymore, the array can be zeroed out. This limits available ways of attack on a key as no key material and its artifacts (like passphrase) touch the filesystem and they are not stored. However there are some weaknessess which need to be taken into consideration. First of all, a key derived from passphrase has [its own weaknesses](0x04g-Testing-Cryptography.md#weak-key-generation-functions). Additionally, a key material should be cleared out from memory as soon as it is not need anymore. However, note that some ciphers do not properly clean up their byte-arrays. For instance, the AES Cipher in BouncyCastle does not always clean up its latest working key leaving some copies of the byte-array hanging around. Next, BigInteger based keys (e.g. private keys) cannot be removed from the heap nor zeroed out just like that. Clearing byte array can be achieved by writing a wrapper which implements [Destroyable](https://docs.oracle.com/javase/8/docs/api/javax/security/auth/Destroyable.html#destroy--).
```
sushi2k, line 428: What do you mean with "and construction of passphrase, which is prone to human error"? Would suggest to remove. 


```suggestion
A more common solution (regarding Android API level), however less-user friendly and with some weaknesses is to derive a key from user provided passphrase. This means that the user should be prompted to input a passphrase every time the application needs to perform a cryptographic operation. This is not the ideal implementation from a user point of view. However this approach makes a key available in an array in memory while it is being used and when the key is not needed anymore, the array can be zeroed out. This limits the available ways of attacks on a key as no key material and its artifacts (like a passphrase) touch the filesystem and they are not stored. However there are some weaknesses which need to be taken into consideration. First of all, a key derived from passphrase has [its own weaknesses](0x04g-Testing-Cryptography.md#weak-key-generation-functions). Additionally, the key material should be cleared out from memory as soon as it is not need anymore. However, note that some ciphers do not properly clean up their byte-arrays. For instance, the AES Cipher in BouncyCastle does not always clean up its latest working key leaving some copies of the byte-array in memory. Next, BigInteger based keys (e.g. private keys) cannot be removed from the heap nor zeroed out just like that. Clearing byte array can be achieved by writing a wrapper which implements [Destroyable](https://docs.oracle.com/javase/8/docs/api/javax/security/auth/Destroyable.html#destroy--).
```

Document/0x05d-Testing-Data-Storage.md
```diff
@@ -310,6 +406,36 @@ Inspect the source code to determine whether native Android mechanisms identify
...
+
+A slightly less secure way of storing encryption keys, is in the SharedPreferences of Android. When [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences.html "Android SharedPreference API") are initialized in [MODE_PRIVATE](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE "MODE_PRIVATE"), the file is only readable by the application that created it. However, on rooted devices any other application with root access can simply read the SharedPreference file of other apps, it does not matter whether `MODE_PRIVATE` has been used or not. This is not the case for the AndroidKeyStore. Since AndroidKeyStore access is managed on kernel level, which needs considerably more work and skill to bypass without the AndroidKeyStore clearing or destroying the keys.
```
sushi2k, line 434: Does MODE_PRIVATE anyway have much impact? If the device is not rooted the sharedPref is anyway in the  sandbox, so only the app can access it. If it's rooted MODE_PRIVATE has no relevance anymore. 
So I wouldn't consider this a secure approach. and we should call it "Not recommended" above also if all keys are just stored in SharedPref

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
 
-The list of providers included in Android varies between versions of Android and the OEM-specific builds. Some provider implementations in older versions are now known to be less secure or vulnerable. Thus, Android applications should not only choose the correct algorithms and provide good configuration, in some cases they should also pay attention to the strength of the implementations in the legacy providers.
+KeyStore and KeyChain provide APIs for storing and using keys (behind the scene, KeyChain API uses KeyStore system). These systems allow to administer the full lifecycle of the cryptographic keys. Requirements and guidance for implementation of cryptographic key management can be found in [Key Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Key_Management_Cheat_Sheet.html "Key Management Cheat Sheet"). We can identify following phases:
+
+- generating of a key
```
sushi2k, line 17:
```suggestion
- generating a key
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+- using of a key
```
sushi2k, line 18:
```suggestion
- using a key
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+- storing of a key
```
sushi2k, line 19:
```suggestion
- storing a key
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+- archiving of a key
```
sushi2k, line 20:
```suggestion
- archiving a key
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+- deleting of a key
```
sushi2k, line 21:
```suggestion
- deleting a key
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+
+> Please note that storing of a key is analyzed in the chapter "[Testing Data Storage](0x05d-Testing-Data-Storage.md)".
+
+These phases are managed by Keystore/KeyChain system. However how the system works depends on how application developer used it. Then, you should focus on functions which are used by application developer. You should identify and verify following functions:
```
sushi2k, line 25:
```suggestion
These phases are managed by the Keystore/KeyChain system. However how the system works depends on how the application developer implemented it. For the analysis process you should focus on functions which are used by the application developer. You should identify and verify the following functions:
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -1,16 +1,69 @@
...
+  - Sockets have changed from `OpenSSLSocketImpl` to `ConscryptFileDescriptorSocket`, and `ConscryptEngineSocket`.
+  - `SSLSession` with null parameters give a `NullPointerException`.
+  - You need to have large enough arrays as input bytes for generating a key otherwise, an `InvalidKeySpecException` is thrown.
+  - If a Socket read is interrupted, you get a `SocketException`.
+- For Android 9 (API level 28) and above the [Android Developer Blog](https://android-developers.googleblog.com/2018/03/cryptography-changes-in-android-p.html "Cryptography Changes in Android P") shows even more changes:
+  - You get a warning if you still specify a security provider using the `getInstance` method and you target any API below 28. If you target Android 9 (API level 28) or above, you get an error.
+  - The `Crypto` security provider is now removed. Calling it will result in a `NoSuchProviderException`.
+- For Android 10 (API level 29) the [Developer Documentation](https://developer.android.com/about/versions/10/behavior-changes-all#security "Security Changes in Android 10") list all network security changes.
```
sushi2k, line 46:
```suggestion
- For Android 10 (API level 29) the [Developer Documentation](https://developer.android.com/about/versions/10/behavior-changes-all#security "Security Changes in Android 10") lists all network security changes.
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -150,264 +205,211 @@ KeyPair keyPair = keyPairGenerator.generateKeyPair();
...
+}
+```
 
-- `Cipher`
-- `Mac`
-- `MessageDigest`
-- `Signature`
-- `Key`, `PrivateKey`, `PublicKey`, `SecretKey`
-- And a few others in the `java.security.*` and `javax.crypto.*` packages.
+The above method requires a character array containing the password and the needed key length in bits, for instance a 128 or 256-bit AES key. We define an iteration count of 10000 rounds which will be used by the PBKDF2 algorithm. Increasing number of iteration significantly increases the workload for a brute-force attack on password, however it can affect performance as more computational power is required for key derivation. We define the salt size equal to the key length, we divide by 8 to take care of the bit to byte conversion. We use the `SecureRandom` class to randomly generate a salt. Obviously, the salt is something you want to keep constant to ensure the same encryption key is generated time after time for the same supplied password. Note that you can store the salt privately in `SharedPreferences`. It is recommended to exclude the salt from the Android backup mechanism to prevent synchronization in case of higher risk data.
```
sushi2k, line 227:
```suggestion
The above method requires a character array containing the password and the needed key length in bits, for instance a 128 or 256-bit AES key. We define an iteration count of 10,000 rounds which will be used by the PBKDF2 algorithm. Increasing number of iteration significantly increases the workload for a brute-force attack on password, however it can affect performance as more computational power is required for key derivation. We define the salt size equal to the key length, we divide by 8 to take care of the bit to byte conversion. We use the `SecureRandom` class to randomly generate a salt. Obviously, the salt is something you want to keep constant to ensure the same encryption key is generated time after time for the same supplied password. Note that you can store the salt privately in `SharedPreferences`. It is recommended to exclude the salt from the Android backup mechanism to prevent synchronization in case of higher risk data.
```

Document/0x05e-Testing-Cryptography.md
```diff
@@ -150,264 +205,211 @@ KeyPair keyPair = keyPairGenerator.generateKeyPair();
...
 
-#### Overview
+#### Random number generation
 
-Cryptography requires secure pseudo random number generation (PRNG). Standard Java classes do not provide sufficient randomness and in fact may make it possible for an attacker to guess the next value that will be generated, and use this guess to impersonate another user or access sensitive information.
+Cryptography requires secure pseudo random number generation (PRNG). Standard Java class (means java.util.Random) does not provide sufficient randomness and in fact may make it possible for an attacker to guess the next value that will be generated, and use this guess to impersonate another user or access sensitive information.
```
sushi2k, line 235:
```suggestion
Cryptography requires secure pseudo random number generation (PRNG). Standard Java classes as `java.util.Random` do not provide sufficient randomness and in fact may make it possible for an attacker to guess the next value that will be generated, and use this guess to impersonate another user or access sensitive information.
```

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: By prone to human error I mean passphrase which is reused, easy to guess and other flaws which are related to password created by human. Do you mean to remove that 'prone to human error' or whole section?

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: all these 4 points (from key derived from passphrase to all keys in SharedPreferences) or only 'all keys in SharedPreferences'?

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: I will update this. However as it depends on one of previous comments I will update them both.

Document/0x05d-Testing-Data-Storage.md
sushi2k: "all keys are stored in SharedPreferences" shouldn't be recommended. 
As long as keys are in the KeyStore then it's best practice, as we are suggesting to leverage a strong encryption mechanisms offered by  Android.

Document/0x05d-Testing-Data-Storage.md
sushi2k: Got it. I rephrased my suggestion in this part so it's more clearer. Please have a look.

Document/0x05d-Testing-Data-Storage.md
sushi2k: Yes, this goes back to the suggestion of marking it "not recommend" in the section above.

sushi2k:
@lwierzbicki please have a look at the answers and then we are ready to go :-) Thanks for the really great work you did!

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: Done. :)

Document/0x05d-Testing-Data-Storage.md
lwierzbicki: Trying to do it as simple as possible. Have a look :)

PR #1720, opened 2020-05-07
Add two files associated with jailbreaks

martinbydefault:
There is this [list of files associated with jailbreaks](https://www.trustwave.com/en-us/resources/blogs/spiderlabs-blog/jailbreak-detection-methods/) from trustwave that has two additional files to check:

    /var/tmp/cydia.log
    /var/log/syslog

Here is more evidence regarding /var/log/syslog is related to jailbreak tools: https://www.theiphonewiki.com/wiki/System_Log#On-device_with_saving_to_a_file

I haven't found an alternative supporting evidence for the /var/tmp/cydia.log file, I'm trusting trustwave reference in this case.

Feel free to discuss, and of course I can remove this file from the commit if there isn't enough information supporting it's related to jailbreak.


martinbydefault:
I didn't noticed at first, but the trustwave article I'm mentioning is referenced in this same file.

Issue #1736, opened 2020-05-14
Migrate to new github actions


commjoen:
In order to speed up reviewing & releasing, we need to :

- [ ] Fix the epub rendering (check document for links and images) and fix errors from `epubcheck OWASP_MSTG-SNAPSHOT-en.epub` 
- [ ] Fix the mobi rendering warnings/errors (missing images, links)
- [ ] Cleanup old materials
- [ ] Fix the PDF generation:
   - [ ] Reduce the TOC by one level, at the moment the TOC is 20 pages / TOC should have one or two levels less: `--toc-depth=3` not working right now.
   - [ ] `--number-sections` not taking effect
   - [ ] Look into Latex errors: #1772 (not a priority)
   - [ ] Fix links inside of the pdf:
      - [x] Broken links
      - [ ] Internal links broken (at least in Mac OS X Preview app, surely in other readers)
   - [x] Fix page layouts (new page and line breaks! Over and underflows) - SHOULD BE DONE LAST!
      - [x] Fix line breaks in problematic strings -> check "Notes" below.
   - [x] Fix cover so it spans over the whole page, at the moment there is a white border
   - [x] Fix images (make sure all are rendered: still need to fix some sizes & a few missing images);  page 657 and 845 pictures are missing @sushi2k   will look into it ([WARNING] Could not fetch resource 'Images/Chapters/0x5d/Android9_secure_key_import_to_keystore.png': PandocResourceNotFound "Images/Chapters/0x5d/Android9_secure_key_import_to_keystore.png")
   - [x] Level 6 headings (######) have no formatting and look like normal text. General Testing Guide, Android and iOS level 1 will be removed.  Generic / iOS and Android becomes a prefix of the sections. Need to consolidate also the wording of the titles (https://github.com/OWASP/owasp-mstg/pull/1775)
   - [x] Code is not properly highlighted. It just has another font and size, but no background or box around it: let's make sure always a supported language is chosen OR a new lib to support more langs.
- [x] Create new version of Docker Image as Syntax highlighting was added
- [x] Create MSTG generator container and push it to OWASP. registry at docker hub
- [x] Fix OWASP Logo, use new one from MASVS (being fixed in #1775)
- [x] Ensure consistency of chapter names (being fixed in #1775)
- [x] Remove empty files (being fixed in #1775)
- [x] Include Foreword (being fixed in #1775)
[...]

PR #1775, opened 2020-06-12
Removing first levels in sections

cpholguera:
This PR supports #1736 .


cpholguera:
@sushi2k This is ready for review now. As we discussed, let's leave the issue with the in-document links for a new PR and merge this one if we're already happy with the results. As @commjoen pointed out, let's make sure we announce after the merge so that everybody updates their current PRs as this is quite _invasive_.

PR #1809, opened 2020-08-23
Enhancing URL loading

cpholguera:
- Extended Intro regarding _trusted content_
- Added _handling page navigation_ to "static analysis"
- Added "dynamic analysis" and references


Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -524,9 +524,9 @@ The [Vulnerable App](https://github.com/clviper/android-fragment-injection/raw/m
...
+One of the most important things to do when testing WebViews is to make sure that only trusted content can be loaded in it. Any newly loaded page could be potentially malicious, try to exploit any WebView bindings or try to phish the user. Unless you're developing a browser app, usually you'd like to restrict the pages being loaded to the domain of your app. A good practice is to prevent the user from even having the chance to input any URLs inside WebViews (which is the default on Android) nor navigate outside the trusted domains, even when navigating on trusted domains there's still the risk that the user might encounter and click on other links to untrustworthy content (e.g. if the page allows for other users to post comments for instance. In addition, some developers might even override some default behaviour which can be potentially dangerous for the user, see more details below in the Static Analysis section.
```
TheDauntless, line 529:
```suggestion
One of the most important things to do when testing WebViews is to make sure that only trusted content can be loaded in it. Any newly loaded page could be potentially malicious, try to exploit any WebView bindings or try to phish the user. Unless you're developing a browser app, usually you'd like to restrict the pages being loaded to the domain of your app. A good practice is to prevent the user from even having the chance to input any URLs inside WebViews (which is the default on Android) nor navigate outside the trusted domains. Even when navigating on trusted domains there's still the risk that the user might encounter and click on other links to untrustworthy content (e.g. if the page allows for other users to post comments). In addition, some developers might even override some default behavior which can be potentially dangerous for the user. See the Static Analysis section below for more details.
```

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -544,6 +544,19 @@ Virus Total provides an API for analyzing URLs and local files for known threats
 
 ### Static Analysis
 
+As we mentioned before, [handling page navigation](https://developer.android.com/guide/webapps/webview#HandlingNavigation "Handling page navigation") should be analyzed carefully, especially when you suspect that the user might be able to open any external pages. The default and the safest behaviour on Android is to the default web browser open any link that the user might click. However, the developers might decide not to do this and handle the links within their own WebView by providing a `WebViewClient`. If this is the case, be sure to search for and inspect the following interception callback functions:
```
TheDauntless, line 547:
```suggestion
As we mentioned before, [handling page navigation](https://developer.android.com/guide/webapps/webview#HandlingNavigation "Handling page navigation") should be analyzed carefully, especially when users might be able to navigate away from a trusted environment. The default and safest behavior on Android is to let the default web browser open any link that the user might click inside the WebView. However, this default logic can be modified by configuring a `WebViewClient` which allows navigation requests to be handled by the app itself. If this is the case, be sure to search for and inspect the following interception callback functions:
```
Rewrote it a bit.

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -544,6 +544,19 @@ Virus Total provides an API for analyzing URLs and local files for known threats
...
+
+- `shouldOverrideUrlLoading` allows your application to either abort loading WebViews with suspicious content by returning `true` or allow the WebView to load the URL by returning `false`. Considerations:
+  - This method is not called for POST requests.
+  - This method is not called for XmlHttpRequests, iFrames, "src" attributes included in HTML or `<script>` tags. Instead, `shouldInterceptRequest` should take care of this.
+- `shouldInterceptRequest` allows the application to return the data from resource requests. If the return value is null, the WebView will continue to load the resource as usual. Otherwise, the return response and data will be used. Considerations:
```
TheDauntless, line 552:
```suggestion
- `shouldInterceptRequest` allows the application to return the data from resource requests. If the return value is null, the WebView will continue to load the resource as usual. Otherwise, the data returned by the shouldInterceptRequest method is used. Considerations:
```

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -544,6 +544,19 @@ Virus Total provides an API for analyzing URLs and local files for known threats
...
+  - This callback is invoked for a variety of URL schemes (e.g., `http(s):`, `data:`, `file:`, etc.), not only those schemes which send requests over the network.
+  - This is not called for `javascript:` or `blob:` URLs, or for assets accessed via `file:///android_asset/` or `file:///android_res/` URLs.
+In the case of redirects, this is only called for the initial resource URL, not any subsequent redirect URLs.
+  - When Safe Browsing is enabled, these URLs still undergo Safe Browsing checks but the developer can allow the URL with `setSafeBrowsingWhitelist` or even ignore the warning via the `onSafeBrowsingHit` callback.
+  
+As you can see there are a lot of points to consider when testing the security of WebViews having a WebViewClient, be sure to carefully read and understand all of them by checking the [`WebViewClient` Documentation](https://developer.android.com/reference/android/webkit/WebViewClient "WebViewClient").
```
TheDauntless, line 558:
```suggestion
As you can see there are a lot of points to consider when testing the security of WebViews that have a WebViewClient configured, so be sure to carefully read and understand all of them by checking the [`WebViewClient` Documentation](https://developer.android.com/reference/android/webkit/WebViewClient "WebViewClient").
```

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -554,9 +567,12 @@ While the default value of `EnableSafeBrowsing` is `true`, some applications mig
         ...
     </application>
 </manifest>
-
 ```
 
+### Dynamic Analysis
+
+A convenient way to dynamically test deep linking is to use Frida or frida-trace and hook the `shouldOverrideUrlLoading`, `shouldInterceptRequest` methods while using the app and clicking on links within its WebView. Be sure to also hook other related [`Uri`](https://developer.android.com/reference/android/net/Uri "Uri class") methods such as `getHost`, `getScheme` or `getPath` which are typically used to inspect the requests and match known patterns or deny lists.
```
TheDauntless, line 574:
```suggestion
A convenient way to dynamically test deep linking is to use Frida or frida-trace and hook the `shouldOverrideUrlLoading`, `shouldInterceptRequest` methods while using the app and clicking on links within the WebView. Be sure to also hook other related [`Uri`](https://developer.android.com/reference/android/net/Uri "Uri class") methods such as `getHost`, `getScheme` or `getPath` which are typically used to inspect the requests and match known patterns or deny lists.
```

TheDauntless (commented):
Some suggestions and then it's ready to merge :)

sushi2k (approved):
Nice explanation. Thanks

Issue #1363, opened 2019-07-26
5c - Create Tooling (Dynamic Instrumentation)


cpholguera:
Create Tampering and RI -> DI -> Tooling

after "Dynamic Instrumentation"

```
##### Tooling
```

This section will present recommended tools to use for Android dynamic analysis, e.g. Xposed and Frida.

Consider the tools in 5b! Avoid repeating content. Consider if we should move some content here (e.g. Xposed), or for Frida talk about some basics in 5b (actually only what it's being used there) and the more advanced in 5c (e.g. the hooking capabilities).

su-vikas:
will take this one.

PR #1835, opened 2020-10-06
0x05c Android tooling updated

su-vikas:
For issue #1363. 

Added content to Frida and Xposed sub section to Tooling. Most of the details are already covered in the other parts of the guide, just referred to existing content.

### DIFF
--- a/Document/0x05c-Reverse-Engineering-and-Tampering.md
+++ b/Document/0x05c-Reverse-Engineering-and-Tampering.md
@@ -1552,6 +1552,18 @@ $ setprop wrap.com.foo.bar LD_PRELOAD=/data/local/tmp/libpreload.so
 
 ### Dynamic Instrumentation
 
+#### Tooling
+
+##### Frida
+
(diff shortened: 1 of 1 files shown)


cpholguera:
Hi @su-vikas , you're right! And that made me think: the info/overview/intro to the tools is scattered thoughout the guide but we actually have a "Tools" chapter. What if we pack all intros there and simply refer to that section? That has many advantages:
- No repetition.
- One unique place for all tools.
- The Test Cases will only contain "actions" we make using those tools.
- We can even build an overview of where a tools is being used using a parsing script.

What do you think? 

@sushi2k and you?

cpholguera:
UPDATE: I got inspired and I will be opening a PR for this very soon. I'm moving all the tool descriptions to the Tools chapter and will refere back there from the rest of the guide. In that PR I'm removing the "Tooling" section of in 0x6c after "Dynamic Instrumentation". Therefore, we can close this one. But I might need your help @su-vikas when I open the other PR, is that ok for you? Thanks a lot for getting attention to this topic again ^^

cpholguera:
Opened PR: https://github.com/OWASP/owasp-mstg/pull/1843

su-vikas:
@cpholguera  totally agree with this approach. This makes referring to the guide for learning about tools easier too. And sure I can help with the new PR.

PR #1844, opened 2020-10-10
0x05h - Added sms retriever API and declaration forms

### LINKED ISSUE #1777: Add SMS retriever API
Add to https://github.com/OWASP/owasp-mstg/blob/e6ddca17b0a95e521463cf66892a0e4496cd0e7d/Document/0x05h-Testing-Platform-Interaction.md#permission-analysis

With the sms retriever API an app is able to trigger an SMS on the backend and the OS will give it back. It must contain a certain set of characteristics, e.g. a certain hash. This way the app does not have to request read access to SMSs as a permission. Example app: WhatsApp

0x10f2c:
Added in a section regarding replacement APIs for certain permissions (such as the SMS Retriever API), as well as a mention of Declaration forms and how they relate to permission usage

This PR closes #1777


Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -287,7 +287,13 @@ For example if both `READ_EXTERNAL_STORAGE` and `WRITE_EXTERNAL_STORAGE` are lis
 
 ### Permission Analysis
 
-Always check whether the application is requesting permissions it actually needs. Make sure that no permissions are requested which are not related to the goal of the app. For instance: a single-player game that requires access to `android.permission.WRITE_SMS`, might not be a good idea.
+Always check whether the application is requesting permissions it actually requires. Make sure that no permissions are requested which are not related to the goal of the app.
+
+In particular, care should be taken around `DANGEROUS` and `SIGNATURE` permissions as these can affect both the user and the application if mishandled. Take for instance a single-player game that requires access to `android.permission.WRITE_SMS`, this goes against what one would commonly expect from such an application and may indicate potentially malicious usage of the permission.
```
cpholguera, line 292:
```suggestion
Always check whether the application is requesting permissions it actually requires. Make sure that no permissions are requested which are not related to the goal of the app, especially `DANGEROUS` and `SIGNATURE` permissions, since they can affect both the user and the application if mishandled. For instance, it should be suspicious if a single-player game app requires access to `android.permission.WRITE_SMS`.
```

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -287,7 +287,13 @@ For example if both `READ_EXTERNAL_STORAGE` and `WRITE_EXTERNAL_STORAGE` are lis
...
+
+Furthermore declaring the usage of `DANGEROUS` permissions can affect the ability of the application to be hosted on the Google Play Store. The usage of [SMS or Call Log permission groups](https://support.google.com/googleplay/android-developer/answer/9047303?hl=en) are strictly handled when submitting an application to the Play Store, with the application owner having to submit a [Declaration Form](https://support.google.com/googleplay/android-developer/answer/9214102?hl=en) that details the use case for these permissions or risk having their app taken off the Play Store.
+
+Certain `DANGEROUS` permissions also have replacement APIs aimed at streamlining the security and usage of such permissions. One such example is the [SMS Retriever API](https://developers.google.com/identity/sms-retriever/overview) which streamlines the usage of SMS permissions when performing SMS-based user verification. By using this API an application does not have to declare `DANGEROUS` permissions which is a benefit to both the user and application. When analyzing permissions investigate the use case and check to see if there are replacement APIs for the permission in use.
```
cpholguera, line 296: While I personally find that information interesting, we should avoid putting it here since it is very developer focused. We should concentrate on "how to test".

```suggestion
When analyzing permissions, you should investigate the concrete use case scenarios of the app and always check if there are replacement APIs for any `DANGEROUS` permissions in use. A good example is the [SMS Retriever API](https://developers.google.com/identity/sms-retriever/overview) which streamlines the usage of SMS permissions when performing SMS-based user verification. By using this API an application does not have to declare `DANGEROUS` permissions which is a benefit to both the user and application.
```

cpholguera (changes requested):
@0x10f2c This is great, thanks for the new information! Please take a look at the suggestions and comments.

0x10f2c:
Hey I've committed these changes however I think its worth having a one sentence mention of the declaration forms even its just something like "Note usage of these APIS require declaration forms " as it helps inform the tester about restrictions around these APIs and provides further rational to why these APIs should not be used unless completely necessary.

Document/0x05h-Testing-Platform-Interaction.md
```diff
@@ -287,7 +287,9 @@ For example if both `READ_EXTERNAL_STORAGE` and `WRITE_EXTERNAL_STORAGE` are lis
...
+Always check whether the application is requesting permissions it actually requires. Make sure that no permissions are requested which are not related to the goal of the app, especially `DANGEROUS` and `SIGNATURE` permissions, since they can affect both the user and the application if mishandled. For instance, it should be suspicious if a single-player game app requires access to `android.permission.WRITE_SMS`.
+
+When analyzing permissions, you should investigate the concrete use case scenarios of the app and always check if there are replacement APIs for any `DANGEROUS` permissions in use. A good example is the [SMS Retriever API](https://developers.google.com/identity/sms-retriever/overview) which streamlines the usage of SMS permissions when performing SMS-based user verification. By using this API an application does not have to declare `DANGEROUS` permissions which is a benefit to both the user and application.
```
cpholguera, line 292: @0x10f2c  I agree with your comment, what about this?

```suggestion
When analyzing permissions, you should investigate the concrete use case scenarios of the app and always check if there are replacement APIs for any `DANGEROUS` permissions in use. A good example is the [SMS Retriever API](https://developers.google.com/identity/sms-retriever/overview) which streamlines the usage of SMS permissions when performing SMS-based user verification. By using this API an application does not have to declare `DANGEROUS` permissions which is a benefit to both the user and developers of the application, who doesn't have to submit the [Permissions Declaration Form](https://support.google.com/googleplay/android-developer/answer/9214102?hl=en).
```

cpholguera (changes requested):
Please take a look, this might do it. Else please feel free to suggest sth. different ;)

cpholguera (approved):
Thank you @0x10f2c ! :)

Issue #1787, opened 2020-07-09
React Native and Flutter are not hybrid as per the given definition


commjoen:
From https://github.com/OWASP/owasp-masvs/issues/461  by @mrrenaud

https://github.com/OWASP/owasp-mstg/blob/1.1.3/Document/0x04a-Mobile-App-Taxonomy.md#hybrid-app

Hybrid apps are described this way :

A hybrid app executes like a native app, but a majority of the processes rely on web technologies, meaning a portion of the app runs in an embedded web browser (commonly called “webview”)

But this is not true for framework such as React Native, Flutter or even Xamarin. They may have 3 different approaches, but none of them uses an embedded browser.

Do you think it would be interesting to clarify that point, and maybe introduce a new category?

jadeboer:
Absolutely necessary.

su-vikas:
i will take this one.

vixentael:
Shouldn't this be closed?

cpholguera:
That's right, thanks @vixentael for keeping an eye on the issues ;)

PR #1851, opened 2020-10-27
Including Xamarin, React Native & Flutter in Native apps

su-vikas:
For #1787 

I propose to include Xamarin, Flutter and React Native apps as Native apps. Some reasons being: 

- All these frameworks claim to develop "native apps", using Native APIs and with native performance.  [Xamarin](https://dotnet.microsoft.com/apps/xamarin) website is full of word "native". 
- The output is unlike web apps (the other broader category to compare too) and more like a native app. 
- EU study  on  "[Security Privacy and data protection in mobile applications](https://www.enisa.europa.eu/publications/privacy-and-data-protection-in-mobile-applications/at_download/fullReport)" Pg. 32 categorizes the applications into 3 as well.  Thus we can stick with categories used by other major organizations. 
- Adding another category (lets say "nearly native" or "almost native") doesn't add any considerable value and will only lead to complicating and re-adapting some of our existing content. 

Instead if needed  we may add a note - In the book when we use "native" it mostly refer to apps developed using official languages, as we dont discuss about other frameworks anyways. 

Let me know if this sounds good?

### DIFF
--- a/Document/0x04a-Mobile-App-Taxonomy.md
+++ b/Document/0x04a-Mobile-App-Taxonomy.md
@@ -14,7 +14,13 @@ Native apps inherently have the capability to provide the fastest performance wi
 
 Some ambiguity exists when discussing *native apps* for Android as the platform provides two development kits - the Android SDK and the Android NDK. The SDK, which is based on the Java and Kotlin programming language, is the default for developing apps. The NDK (or Native Development Kit) is a C/C++ development kit used for developing binary libraries that can directly access lower level APIs (such as OpenGL). These libraries can be included in regular apps built with the SDK. Therefore, we say that Android *native apps* (i.e. built with the SDK) may have *native* code built with the NDK.
 
-The most obvious downside of *native apps* is that they target only one specific platform. To build the same app for both Android and iOS, one needs to maintain two independent code bases, or introduce often complex development tools to port a single code base to two platforms (e.g. [Xamarin](https://www.xamarin.com/ "Xamarin")).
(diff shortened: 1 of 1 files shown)


Document/0x04a-Mobile-App-Taxonomy.md
```diff
@@ -14,7 +14,13 @@ Native apps inherently have the capability to provide the fastest performance wi
...
 
-The most obvious downside of *native apps* is that they target only one specific platform. To build the same app for both Android and iOS, one needs to maintain two independent code bases, or introduce often complex development tools to port a single code base to two platforms (e.g. [Xamarin](https://www.xamarin.com/ "Xamarin")).
+The most obvious downside of *native apps* is that they target only one specific platform. To build the same app for both Android and iOS, one needs to maintain two independent code bases, or introduce often complex development tools to port a single code base to two platforms. Currently there are following frameworks which offer such functionality:
```
TheDauntless, line 17:
```suggestion
The most obvious downside of *native apps* is that they target only one specific platform. To build the same app for both Android and iOS, one needs to maintain two independent code bases, or introduce often complex development tools to port a single code base to two platforms. The following frameworks are an example of the latter and allow you to compile a single codebase for both Android and iOS.
```
Rewrote the last sentence a few times, and ended up with this.

TheDauntless (changes requested):

TheDauntless:
Thanks for your well argumented PR!

It's a bit weird to call them native apps (especially Xamarin, which uses a native runtime to run non-native code), but they're definitely not at their place in the hybrid section, so this is better 👍

cpholguera (approved):
thanks @su-vikas ;)

TheDauntless (approved):

PR #1870, opened 2021-01-19
Add Websec.ca to list of users

Xar1c:
Websec had been on the list of users of the MSTG but was removed, citing a dead link. This is a request to re-add Websec to the list of users.

This PR closes #1748.


cpholguera (approved):
Hi @Xar1c, we've reviewed what happened and we're very sorry that it was removed because of being "dead" at the time of that PR review. We're happy to add it again :)

PR #1888, opened 2021-03-30
New way to implement cert pinning for iOS

julepka:
Added details about a new approach to implementing certificate pinning on iOS. As per Apple article https://developer.apple.com/news/?id=g9ejcf8y it is now possible to specify CA public key in Info.plist file, and this should be enough for pinning to work.

Also, replaced the code example with a reference to the official documentation and simplified the description.
___
___
This contribution was discussed with @sushi2k and @vixentael

### DIFF
--- a/Document/0x06g-Testing-Network-Communication.md
+++ b/Document/0x06g-Testing-Network-Communication.md
@@ -222,7 +222,9 @@ Verify that the server certificate is pinned. Pinning can be implemented on vari
 2. Limiting certificate issuer to e.g. one entity and bundling the intermediate CA's public key into the application. In this way we limit the attack surface and have a valid certificate.
 3. Owning and managing your own PKI. The application would contain the intermediate CA's public key. This avoids updating the application every time you change the certificate on the server, due to e.g. expiration. Note that using your own CA would cause the certificate to be self-singed.
 
-The code presented below shows how it is possible to check if the certificate provided by the server matches the certificate stored in the app. The method below implements the connection authentication and tells the delegate that the connection will send a request for an authentication challenge.
+The latest approach recommended by Apple is to specify a pinned CA public key in the `Info.plist` file under App Transport Security Settings. You can find an example in their article [Identity Pinning: How to configure server certificates for your app](https://developer.apple.com/news/?id=g9ejcf8y "Identity Pinning: How to configure server certificates for your app").
+
+Another common approach is to check if the certificate provided by the server matches the certificate stored in the app. The method below implements the connection authentication and tells the delegate that the connection will send a request for an authentication challenge.
 
 The delegate must implement `connection:canAuthenticateAgainstProtectionSpace:` and `connection: forAuthenticationChallenge`. Within `connection: forAuthenticationChallenge`, the delegate must call `SecTrustEvaluate` to perform customary X.509 checks. The snippet below implements a check of the certificate.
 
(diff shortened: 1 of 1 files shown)


Document/0x06g-Testing-Network-Communication.md
```diff
@@ -222,7 +222,9 @@ Verify that the server certificate is pinned. Pinning can be implemented on vari
...
+The latest approach recommended by Apple is to specify a pinned CA public key in the `Info.plist` file under App Transport Security Settings. You can find an example in their article [Identity Pinning: How to configure server certificates for your app](https://developer.apple.com/news/?id=g9ejcf8y "Identity Pinning: How to configure server certificates for your app").
+
+Another common approach is to check if the certificate provided by the server matches the certificate stored in the app. The method below implements the connection authentication and tells the delegate that the connection will send a request for an authentication challenge.
 
 The delegate must implement `connection:canAuthenticateAgainstProtectionSpace:` and `connection: forAuthenticationChallenge`. Within `connection: forAuthenticationChallenge`, the delegate must call `SecTrustEvaluate` to perform customary X.509 checks. The snippet below implements a check of the certificate.
```
cpholguera, line 229: do you think that we still need the code below? what if we remove it since we also say "Note that the certificate pinning example above has a major drawback when you use certificate pinning and the certificate changes".

The article that you're referencing has info regarding this. If we link to thst section in the article it should be sufficient.
We're trying to make the MSTG more maintainable, I think this is a good chance to address that. It would be great if you can help us with that. One approach is by identifying things like this that can be found in official sources and simply linking to them instead of maintain explanations and code. Thanks a lot in advance!

cpholguera (changes requested):
Very nice addition! Nothing to object 😌 I only have one request, could you please consider the comment below? That'd be very helpful, thanks a lot!

Document/0x06g-Testing-Network-Communication.md
julepka: Yeah, I agree. I've heard that MSTG is moving away from using code examples because they are not easy to maintain, especially when there are official sources with similar examples.
I'm going to update it right now.

julepka: Also, I think that we still need to mention this approach (without the detailed code example) because it is one of the ways pinning can be implemented. This is a static analysis section, so it looks useful to me to mention different ways of implementation.

julepka:
@cpholguera Updated. Let me know if you have any other suggestions. Thanks!

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -222,40 +222,17 @@ Verify that the server certificate is pinned. Pinning can be implemented on vari
...
-  The control below can verify if the certificate received by the server is matching the one pinned in the client.
-  if ([remoteCertificateData isEqualToData:localCertData]) {
-  NSURLCredential *credential = [NSURLCredential credentialForTrust:serverTrust];
-  [[challenge sender] useCredential:credential forAuthenticationChallenge:challenge];
-}
-else {
-  [[challenge sender] cancelAuthenticationChallenge:challenge];
-}
-```
+Note that if you compare local and remote certificates, you will have to update the app when the remote certificate changes. A fallback certificate can be stored in the app to make the transition more smooth. Alternatively, the pin can be based on public-key comparison. Thus if the remote certificate changes, the public key stays the same.
```
cpholguera, line 229:
```suggestion
Note that if you compare local and remote certificates, you will have to update the app when the remote certificate changes. A fallback certificate can be stored in the app to make the transition smoother. Alternatively, the pin can be based on public-key comparison. Thus if the remote certificate changes, the public key stays the same.
```

cpholguera (approved):
That's perfect, thanks a lot again @julepka!

cpholguera:
Thanks for offering to take care of fixing that, it's very much appreciated. It seems that the link-checker is not really stable. Until we have a better solution you can try to run the link-checker job again and see if it works fine :)

PR #1897, opened 2021-04-10
Update style_guide.md regarding Markdown Link Titles

### LINKED ISSUE #1896: [Style] Remove Title Enforcement for Markdown links
The link style we're using is not maintainable, we have to manually review this for each PR instead of letting the linter do this.

```
For web links, use the normal markdown in-line link format: [TEXT](URL "NAME")
```

There's no rule for title in markdown links:
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md

I don't see how this brings any value, therefore I suggest we remove it.

cpholguera:
This PR closes #1896  .

### DIFF
--- a/style_guide.md
+++ b/style_guide.md
@@ -201,29 +201,29 @@ Throughout the guide, you may want to address the readers in order to tell them 
 
 ### Web Links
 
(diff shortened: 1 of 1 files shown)


TheDauntless (approved):
🐼

PR #1920, opened 2021-07-04
Wrong exploit on Testing for Fragment Injection

MrDoel:
In the exploit.apk, the code is, i.setData(Uri.parse("https://security.claudio.pt")); not Intent intent = xxx


sushi2k (approved):
Nice spot, makes sense. Thank you

PR #1925, opened 2021-07-13
updated iOS binary protection checks

### LINKED ISSUE #1894: Review and enhance iOS Binary Protections
## Task 
enhance the following Test Case:

https://github.com/OWASP/owasp-mstg/blob/master/Document/0x06i-Testing-Code-Quality-and-Build-Settings.md#make-sure-that-free-security-features-are-activated-mstg-code-9

using this article:

https://sensepost.com/blog/2021/on-ios-binary-protections/

## Summary

The decision regarding which protections can and should be enabled is heavily influenced based on if Swiftlang is involved and whether the target binary is an executable or a library.

- PIE – Add the -fPIC compiler flag to the projects build settings. This will only be applicable to the main executable.
- ARC – This will be automatically enabled for Swift only projects (via the swiftc compiler), and added by setting YES to the Objective-C Automatic Reference Counting section in the projects configuration.
- Canary – Enabled by adding the -fstack-protector-all compiler flag to Objective-C projects. If Swift is involved its possible to have it enabled when the library is a hybrid of Objective-C and Swift, but it could show as disabled which is okay.
- Both static and dynamic analysis are possible

su-vikas:
Updated the iOS binary protection checks part. 

This PR closes #1894

### DIFF
--- a/Document/0x06i-Testing-Code-Quality-and-Build-Settings.md
+++ b/Document/0x06i-Testing-Code-Quality-and-Build-Settings.md
@@ -613,20 +613,24 @@ There are various well written explanations which can help with taking care of m
 Although Xcode enables all binary security features by default, it may be relevant to verify this for an old application or to check for the misconfiguration of compilation options. The following features are applicable:
 
 - **ARC** - Automatic Reference Counting - A memory management feature that adds retain and release messages when required
-- **Stack Canary** - Helps prevent buffer overflow attacks by means of having a small integer right before the return pointer. A buffer overflow attack often overwrites a region of memory in order to overwrite the return pointer and take over the process-control. In that case, the canary gets overwritten as well. Therefore, the value of the canary is always checked to make sure it has not changed before a routine uses the return pointer on the stack.
-- **PIE** - Position Independent Executable - enables full ASLR for binary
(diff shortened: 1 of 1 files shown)


Document/0x06i-Testing-Code-Quality-and-Build-Settings.md
```diff
@@ -695,9 +703,21 @@ Below are procedures for checking the binary security features described above.
     [SNIP]
     ```
 
+    This check works for all cases, including pure Swift apps.
+
 ### Dynamic Analysis
 
-Dynamic analysis is not applicable for finding security features offered by the toolchain.
+Dynamically these checks can be performed using Objection tool. One such example is below:
```
cpholguera, line 710:
```suggestion
hese checks can be performed dynamically using Objection. Here's one example:
```

please add the link to the Tools chapter for #objection

cpholguera (changes requested):
Thanks @su-vikas, please take a look at the suggestions and comments. Sorry for the late review, as we communicated we have many topics still open for refactoring the MSTG and MASVS so it's very challenging to keep up with the PRs. We're very sorry for the delays!

PR #1926, opened 2021-07-13
Release v1.2

sushi2k:
This PR will create our new release v1.2 by using our new document build pipeline. 

I shifted everything that is not needed as of now to `tools/docker/old` and `tools/old`, we would need to remove it once we have build the release successfully. 

Image tags were replaced with markdown syntax. 

Created a Python script that creates contributors and mini-contributors dynamically from the Github API. We can use this output to create the list with the GitHub handles automatically during release of the document creation. 

To-dos for release:
https://github.com/OWASP/owasp-mstg/issues/1911


cpholguera:
Thanks Sven, it's already looking very nice I think!

I've found the following overflows for now:

```

This abstract class parses the extra data fields of an Intent, in particular, the
PreferenceActivity.EXTRA_SHOW_FRAGMENT(:android:show_fragment) and PreferenceActivity.
EXTRA_SHOW_FRAGMENT_ARGUMENTS(:android:show_fragment_arguments) fields.

Another common approach is to use the connection:willSendRequestForAuthenticationChallenge:
method of NSURLConnectio

Hook the method we have seen in the static analysis (init(activityItems:applicationActivities:))
to get the activityItems and applicationActivities

For the text file there are two different activity items and “com.apple.UIKit.activity.MarkupAsPDF”
is excluded.

Check for other data types (typically UTIs) with containsPasteboardTypes:inItemSet:.
You may inspect for more concrete data types like, for example an picture as public.
png and public.tiff (UTIs) or for custom data such as com.mycompany.myapp.mytype.
Remember that, in this case, only those apps that declare knowledge of the type

```

sushi2k:
Thanks Carlos! Those are now fixed (put in white spaces in between to get the line break)

cpholguera:
Mor to be fixed until page 182:
- Missing OWASP logo in frontispiece
- Missing "and" and final "full stop" in mini contributors
- "The" should not be uppercase. Search for: "Note that The efficiency of a cipher suite depends on the efficiency of its algorithms."
- Bullet point titles should be bold in "Memory Corruption Bugs (MSTG-CODE-8)"

sushi2k:
Thanks Carlos! Everything fixed you identified.

Document/CHANGELOG.md
```diff
@@ -2,14 +2,17 @@
 
 This document is automatically generated at {{gitbook.time}}
```
cpholguera, line 3:
```suggestion
This document is automatically generated.
```

CHANGELOG.md
```diff
@@ -2,14 +2,17 @@
 
 This document is automatically generated at {{gitbook.time}}
```
cpholguera, line 3:
```suggestion
This document is automatically generated.
```

cpholguera (approved):
Thanks Sven! Looking AWESOME 😃

PR #1958, opened 2021-10-29
Update mlc_config.json

cpholguera:
allow only code 200 for alive

### DIFF
--- a/.github/workflows/checkLinks.yml
+++ b/.github/workflows/checkLinks.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           fetch-depth: 1
       - name: link-check
-        uses: gaurav-nelson/github-action-markdown-link-check@820c0581641b6e2fb34ae4d5f93ef4d841fb37e6
+        uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
           use-quiet-mode: 'yes'
           use-verbose-mode: 'yes'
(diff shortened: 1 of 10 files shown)


cpholguera:
any ideas on what to do with the 403s @anantshri?

anantshri:
@cpholguera my suggestion would be to keep adding domains in the ignore pattern for now. till the list is now growing above 100 entries. 

If it does that cross it then we think of something else. a possible logic could be look for 403 with cloudflare in text then its okey if not then raise a flag.

cpholguera:
ok. Thanks for the feedback @anantshri . But we really need to come up with a solution for this. It's very disturbing for contributors and for us always having to explain that it's ok that this check fails.

I'll enable the option to only check changed files for PRs. This should improve the experience, if someone's introducing a new dead link it will be reported. If it's not dead, the contributors have to add the exceptions themselves to this list. This will be of course documented.

For now I'll add these to the list:

https://www.verticalstructure.com
 https://support.virustotal.com/hc/en-us/articles/115002146549-Mobile-Apps

anantshri:
Just a note https://github.com/tcort/markdown-link-check/issues/109 is the issue in markdown-link-check which should sort this out eventually.

cpholguera:
Great, thanks for checking that @anantshri, very helpful!

sushi2k (approved):

sushi2k:
Thank you for the workaround @cpholguera and @anantshri!

PR #1979, opened 2021-11-16
Remove unnecessary backslash

TheDauntless:
Very tiny fix

### DIFF
--- a/Document/0x06h-Testing-Platform-Interaction.md
+++ b/Document/0x06h-Testing-Platform-Interaction.md
@@ -129,7 +129,7 @@ If having the original source code, you can verify the permissions included in t
 
 You may switch the view to display the raw values by right-clicking and selecting "Show Raw Keys/Values" (this way for example `"Privacy - Location When In Use Usage Description"` will turn into `NSLocationWhenInUseUsageDescription`).
 
-![OWASP MSTG](Images/Chapters/0x06h/purpose_strings_xcode.png) \
+![OWASP MSTG](Images/Chapters/0x06h/purpose_strings_xcode.png)
 
 If only having the IPA:
 


sushi2k (commented):
We are using this throughout the guide it seems (with the backslash at the end) for images. Unfortunately I cannot recall why. Usually this would be a line break. Let's keep it for now and raise a ticket, as this might be some workaround for the document build pipeline

cpholguera:
@sushi2k I found the commit where those backslashes where added: https://github.com/OWASP/owasp-mstg/blob/707b1f7b2a0f516ab76be3607c355daf85b647a3/Document/0x02-Frontispiece.md

"fixed issue with captions in epub and docx"

Do you recall what was happening without them?

cpholguera:
I assume this is why:

https://pandoc.org/MANUAL.html#images

See "Extension: implicit_figures"

```md
If you just want a regular inline image, just make sure it is not the only thing in the paragraph. One way to do this is to insert a nonbreaking space after the image:

![This image won't be a figure](/url/of/image.png)\
```

We still have the issue that the images do not have a `width` so they're being shown at maximum size. Maybe we can include imgs as HTML:

```html
<img href="file.jpg" style="width: 50%;" alt="some caption"/>
```

and, right before giving it to pandocker, we have a short script converting those to:

```md
![some caption](file.jpg){ width=50% }
```

Having HTML imgs in the original document means that markdown renderers will render them properly including width. And then pandocker will take care of converting the markdown to whatever other format because we use the official supported syntax from pandoc markdown extension.

cpholguera:
@TheDauntless see my last comment, I've introduced that in https://github.com/OWASP/owasp-mstg/pull/2059

tools/docker/pandoc_makedocs.sh will run tools/docker/imagereplace.sed so we can input HTML imgs now.

So maybe this PR should remove all those backslashes and convert the images to HTML including a width. For this we have to review the PDF, since some images will be ok but others are for instance extremely big.

PR #2032, opened 2022-01-13
there is no "link above"

Brasco:
The "link above" was an article referred in the owasp cheat sheet

- [ ] Your contribution is written in the 2nd person (e.g. you)
- [ ] Your contribution is written in an active present form for as much as possible.
- [ ] Your contribution has proper formatted markdown and/or code
- [ ] Any references to website have been formatted as [TEXT](URL “NAME”)
- [ ] You verified/tested the effectiveness of your contribution (e.g.: is the code really an effective remediation? Please verify it works!)

This PR closes #< insert number here >.

### DIFF
--- a/Document/0x04e-Testing-Authentication-and-Session-Management.md
+++ b/Document/0x04e-Testing-Authentication-and-Session-Management.md
@@ -380,7 +380,7 @@ Verify that the implementation adheres to JWT [best practices](https://stormpath
 
 #### Enforcing the Hashing Algorithm
 
-An attacker executes this by altering the token and, using the 'none' keyword, changing the signing algorithm to indicate that the integrity of the token has already been verified. Some libraries treated tokens signed with the none algorithm as if they were valid tokens with verified signatures, so the application will trust altered token claims.
+An attacker executes this by altering the token and, using the 'none' keyword, changing the signing algorithm to indicate that the integrity of the token has already been verified. [Some libraries](https://stormpath.com/blog/jwt-the-right-way) might treat tokens signed with the 'none' algorithm as if they were valid tokens with verified signatures, so the application will trust altered token claims.
 
 For example, in Java applications, the expected algorithm should be requested explicitly when creating the verification context:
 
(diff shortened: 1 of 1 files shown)


Document/0x04e-Testing-Authentication-and-Session-Management.md
```diff
@@ -380,7 +380,7 @@ Verify that the implementation adheres to JWT [best practices](https://stormpath
 
 #### Enforcing the Hashing Algorithm
 
-An attacker executes this by altering the token and, using the 'none' keyword, changing the signing algorithm to indicate that the integrity of the token has already been verified. As explained at the link above, some libraries treated tokens signed with the none algorithm as if they were valid tokens with verified signatures, so the application will trust altered token claims.
+An attacker executes this by altering the token and, using the 'none' keyword, changing the signing algorithm to indicate that the integrity of the token has already been verified. Some libraries treated tokens signed with the none algorithm as if they were valid tokens with verified signatures, so the application will trust altered token claims.
```
cpholguera, line 383: Hi @Brasco, thanks for the PR. I'd say the author wanted to refer to this link, which is indeed _above_:

https://stormpath.com/blog/jwt-the-right-way

There you can find this paragraph:

> Always verify the signature before you trust any information in the JWT. This should be a given, but we have recently seen security vulnerabilities in other company’s JWT frameworks. One gotcha that we have seen recently is around the JWT spec that allows you to set signature algorithm to ‘none’. This should be ignored if you expect the JWT to be signed. Put another way, if you are passing a secret signing key to the method that verifies the signature and the signature algorithm is set to ‘none’, it should fail verification.

I suggest the following change. What do you think? If you agree you can simply click on "Commit suggestion", else let me know your thoughts.

```suggestion
An attacker executes this by altering the token and, using the 'none' keyword, changing the signing algorithm to indicate that the integrity of the token has already been verified. [Some libraries](https://stormpath.com/blog/jwt-the-right-way) might treat tokens signed with the 'none' algorithm as if they were valid tokens with verified signatures, so the application will trust altered token claims.
```

cpholguera (changes requested):
Thanks @Brasco, please take a look at my comment.

Document/0x04e-Testing-Authentication-and-Session-Management.md
Brasco: good! thank you!

cpholguera (approved):
Thanks @Brasco!

PR #2042, opened 2022-01-25
Fix Network Security Testing on Android and iOS (by @NowSecure)

cpholguera:
- Both Android and iOS cover now MSTG-NETWORK-1, MSTG-NETWORK-2, MSTG-NETWORK-3, MSTG-NETWORK-4
  - Some Tests were split to cover only one MASVS requirement.
  - NEW: MSTG-NETWORK-1 on both Android and iOS!
- Fix Network Testing chapters structure and cohesion.
- Fix all network cross-chapter references and corrected some content location.

grafik


Document/0x04g-Testing-Cryptography.md
```diff
@@ -44,7 +44,7 @@ When keys need to be transported from one device to another, or from the app to
 
 ### Cryptographic APIs on Android and iOS
 
-While same basic cryptographic principles apply independent of the particular OS, each operating system offers its own implementation and APIs. Platform-specific cryptographic APIs for data storage are covered in greater detail in the "[Data Storage on Android](0x05d-Testing-Data-Storage.md)" and "[Testing Data Storage on iOS](0x06d-Testing-Data-Storage.md)" chapters. Encryption of network traffic, especially Transport Layer Security (TLS), is covered in the "[Android Network APIs](0x05g-Testing-Network-Communication.md)" chapter.
+While same basic cryptographic principles apply independent of the particular OS, each operating system offers its own implementation and APIs. Platform-specific cryptographic APIs for data storage are covered in greater detail in the "[Data Storage on Android](0x05d-Testing-Data-Storage.md)" and "[Testing Data Storage on iOS](0x06d-Testing-Data-Storage.md)" chapters. Encryption of network traffic, especially Transport Layer Security (TLS), is covered in the "[Android Network Communication](0x05g-Testing-Network-Communication.md)" chapter.
```
TheDauntless, line 47:
```suggestion
While same basic cryptographic principles apply independent of the particular OS, each operating system offers its own implementation and APIs. Platform-specific cryptographic APIs for data storage are covered in greater detail in the "[Data Storage on Android](0x05d-Testing-Data-Storage.md)" and "[Testing Data Storage on iOS](0x06d-Testing-Data-Storage.md)" chapters. Encryption of network traffic, especially Transport Layer Security (TLS), is covered in the "[Android Network Communication](0x05g-Testing-Network-Communication.md)" and "[iOS Network Communication](0x06g-Testing-Network-Communication.md)" chapters.
```

Document/0x05c-Reverse-Engineering-and-Tampering.md
```diff
@@ -293,7 +293,7 @@ From here on you can use this information to derive more insights which might be
...
+Remember that in most of the cases, just using static analysis will not be enough and might even turn to be extremely inefficient when compared to the dynamic alternatives which will get much more reliable results (e.g. using an interceptor proxy). In this section we've just slightly touched the surface, please refer to the section "[Basic Network Monitoring/Sniffing](0x05b-Basic-Security_Testing.md#basic-network-monitoringsniffing "Basic Network Monitoring/Sniffing")" in the "Android Basic Security Testing" chapter and also check the test cases in the chapter "[Android Network Communication](0x05g-Testing-Network-Communication.md)".
```
TheDauntless, line 296:
```suggestion
Remember that in most of the cases, just using static analysis will not be enough and might even turn to be extremely inefficient when compared to the dynamic alternatives which will get much more reliable results (e.g. using an interceptor proxy). In this section we've just slightly touched the surface, please refer to the section "[Basic Network Monitoring/Sniffing](0x05b-Basic-Security_Testing.md#basic-network-monitoringsniffing "Basic Network Monitoring/Sniffing")" in the "Android Basic Security Testing" chapter and also check the test cases in the "[Android Network Communication](0x05g-Testing-Network-Communication.md)" chapter.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -15,6 +50,76 @@ Make sure that the hostname and the certificate itself are verified correctly. E
 
 ### Static Analysis
 
+#### Trust Anchors in the Network Security Configuration
+
+When running on Android 7.0 (API level 24) or higher, apps targeting those API levels will use a default Network Security Configuration that doesn't trust any user supplied CAs, reducing the possibility of MITM attacks by luring users to install malicious CAs. However, this protection can be bypassed by using a custom Network Security Configuration with a custom trust anchor indicating that the app will trust user supplied CAs.
```
TheDauntless, line 55:
```suggestion
Applications targeting Android 7.0 (API level 24) or higher will use a default Network Security Configuration that doesn't trust any user supplied CAs, reducing the possibility of MITM attacks by luring users to install malicious CAs. However, this protection can be disabled by using a custom Network Security Configuration with a custom trust anchor indicating that the app will trust user supplied CAs.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -15,6 +50,76 @@ Make sure that the hostname and the certificate itself are verified correctly. E
...
+
+Use a decompiler (e.g. jadx or apktool) to confirm the target SDK version. After decoding the app you can look for the presence of `targetSDK` present in the file apktool.yml that was created in the output folder.
```
TheDauntless, line 57:
```suggestion
Use a decompiler (e.g. jadx or apktool) to confirm the target SDK version. After decoding the app you can look for the presence of `targetSDK` in the file apktool.yml that was created in the output folder.
```

TheDauntless: apktool technically is a disassembler though, not a decompiler.

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -15,6 +50,76 @@ Make sure that the hostname and the certificate itself are verified correctly. E
...
+
+The Network Security Configuration should be analyzed to determine what settings are configured. The file is located inside the APK in the /res/xml/ folder with the name network_security_config.xml.
```
TheDauntless, line 59: That's actually not necessary. You can call it whatever you want. So what about
```suggestion
The Network Security Configuration should be analyzed to determine what settings are configured. The name of the file can be found in the `android:networkSecurityConfig` attribute of the `<application>` tag inside the Android Manifest. The file can then be found in the /res/xml/ folder with the .xml extension.
```
Though it's possible to store it in other folders as well, if you really want to I think.

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -15,6 +50,76 @@ Make sure that the hostname and the certificate itself are verified correctly. E
...
+
+If there are custom `<trust-anchors>` present in a `<base-config>` or `<domain-config>`, that define a `<certificates src="user">` the application will trust user supplied CAs for those particular domains or for all domains. Example:
```
TheDauntless, line 61:
```suggestion
If there are custom `<trust-anchors>` present in a `<base-config>` or `<domain-config>` that define a `<certificates src="user">`, the application will trust user supplied CAs for those particular domains or for all domains. Example:
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -15,6 +50,76 @@ Make sure that the hostname and the certificate itself are verified correctly. E
...
+            <pin digest="SHA-256">YLh1dUR9y6Kja30RrAn7JKnbQG/uEtLMkBgFF2Fuihg=</pin>
+            <!-- Hash of the public key (SubjectPublicKeyInfo of the X.509 certificate) of
+            the Root CA of the OWASP website server certificate -->
+            <pin digest="SHA-256">Vjs8r4z+80wjNcr1YKepWQboSIRi63WsWXhIMN+eWys=</pin>
+        </pin-set>
+    </domain-config>
+</network-security-config>
+```
+
+Is important to understand the precedence of entries. If a value is not set in a `<domain-config\>` entry or in a parent `<domain-config\>`, the configurations in place will be based on the `<base-config\>`, and lastly if not defined in this entry, the default configuration will be used.
```
TheDauntless, line 90:
```suggestion
It is important to understand the precedence of entries. If a value is not set in a `<domain-config>` entry or in a parent `<domain-config\>`, the configurations in place will be based on the `<base-config>`, and lastly, if not defined in this entry, the default configuration will be used.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -93,21 +198,27 @@ Make sure that your application verifies a hostname before setting a trusted con
 
 ### Dynamic Analysis
 
-Dynamic analysis requires an interception proxy. To test improper certificate verification, check the following controls:
+When testing an app targeting Android 7.0 (API level 24) or higher it should be effectively applying the Network Security Configuration **unless the app targets API levels below 24** and you shouldn't able to see the decrypted HTTPS traffic at first.
```
TheDauntless, line 201:
```suggestion
When testing an app targeting Android 7.0 (API level 24) or higher it should be effectively applying the Network Security Configuration and you shouldn't able to see the decrypted HTTPS traffic at first. However,  if the app targets API levels below 24, the app will automatically accept the installed user certificates.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -134,7 +245,7 @@ When only very few pin failures are reported, then the network should be ok, and
 
 ### Static Analysis
 
-#### Network Security Configuration
+#### Certificate Pinning in the Network Security Configuration
 
 To customize their network security settings in a safe, declarative configuration file without modifying app code, applications can use the [Network Security Configuration](https://developer.android.com/training/articles/security-config.html "Network Security Configuration documentation") that Android provides for versions 7.0 and above.
```
TheDauntless, line 250:
```suggestion
To customize their network security settings in a safe, declarative configuration file without modifying app code, applications can use the [Network Security Configuration](https://developer.android.com/training/articles/security-config.html "Network Security Configuration documentation") that Android provides for versions 7.0 (API level 24) and above.
```

Document/0x06c-Reverse-Engineering-and-Tampering.md
```diff
@@ -127,7 +127,7 @@ From here on you can use this information to derive more insights which might be
...
+Remember that in most cases, only using static analysis will not be enough and might even turn to be extremely inefficient when compared to the dynamic alternatives which will get much more reliable results (e.g. using an interceptor proxy). In this section we've just slightly touched the surface, so please refer to the section "[Basic Network Monitoring/Sniffing](0x06b-Basic-Security-Testing.md#basic-network-monitoringsniffing "Basic Network Monitoring/Sniffing")" in the "iOS Basic Security Testing" chapter and check out the test cases in the chapter "[iOS Network Communication](0x06g-Testing-Network-Communication.md)" for further information.
```
TheDauntless, line 130:
```suggestion
Remember that in most cases, using only static analysis will not be enough and might even turn out to be extremely inefficient when compared to the dynamic alternatives which will get much more reliable results (e.g. using an interception proxy). In this section we've only touched the surface, so please refer to the section "[Basic Network Monitoring/Sniffing](0x06b-Basic-Security-Testing.md#basic-network-monitoringsniffing "Basic Network Monitoring/Sniffing")" in the "iOS Basic Security Testing" chapter and check out the test cases in the chapter "[iOS Network Communication](0x06g-Testing-Network-Communication.md)" for further information.
```
Previous update had 'interception proxy'. Haven't checked the MSTG for different versions, but maybe we should choose a version we like.

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -1,29 +1,14 @@
...
+[App Transport Security (ATS)](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html "Information Property List Key Reference: Cocoa Keys") is a set of security checks that the operating system enforces when making connections with [NSURLConnection](https://developer.apple.com/reference/foundation/nsurlconnection "API Reference NSURLConnection"), [NSURLSession](https://developer.apple.com/reference/foundation/urlsession "API Reference NSURLSession") and [CFURL](https://developer.apple.com/reference/corefoundation/cfurl-rd7 "API Reference CFURL") to public hostnames. ATS is enabled by default for applications build on iOS 9 and above.
```
TheDauntless, line 11:
```suggestion
[App Transport Security (ATS)](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html "Information Property List Key Reference: Cocoa Keys") is a set of security checks that the operating system enforces when making connections with [NSURLConnection](https://developer.apple.com/reference/foundation/nsurlconnection "API Reference NSURLConnection"), [NSURLSession](https://developer.apple.com/reference/foundation/urlsession "API Reference NSURLSession") and [CFURL](https://developer.apple.com/reference/corefoundation/cfurl-rd7 "API Reference CFURL") to public hostnames. ATS is enabled by default for applications targeting iOS 9 and above.
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -104,73 +89,76 @@ Starting from January 1 2017, Apple App Store review requires justification if o
...
----
-ATS Default Connection
-Result : PASS
----
+First, you should identify all network requests in the source code and ensure that no plain HTTP URLs are used. Make sure that sensitive information is sent over secure channels by using [`URLSession`](https://developer.apple.com/documentation/foundation/urlsession) (which uses the standard [URL Loading System from iOS](https://developer.apple.com/documentation/foundation/url_loading_system)) or [`Network`](https://developer.apple.com/documentation/network) (for socket-level communication using TLS and access to TCP and UDP).
+
+Next, you should ensure that the app is not allowing cleartext HTTP traffic. Since iOS 9.0 cleartext HTTP traffic is blocked by default (due to App Transport Security (ATS)) but there are multiple ways in which an application can still send it:
+
+- Configuring ATS to enable cleartext traffic by setting the `NSAppTransportSecurity` attribute to `true` on [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) in the app [`Info.plist`](https://developer.apple.com/documentation/bundleresources/information_property_list).
```
TheDauntless, line 116:
```suggestion
- Configuring ATS to enable cleartext traffic by setting the `NSAppTransportSecurity` attribute to `true` on [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) in the app's [`Info.plist`](https://developer.apple.com/documentation/bundleresources/information_property_list).
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -104,73 +89,76 @@ Starting from January 1 2017, Apple App Store review requires justification if o
...
+- Using low-level APIs (e.g. [`Network`](https://developer.apple.com/documentation/network)) or [`CFNetwork`](https://developer.apple.com/documentation/cfnetwork) to set up a custom HTTP connection.
+- Using a cross-platform framework (e.g. Flutter, Xamarin, ...), as these typically have their own implementations for HTTP libraries.
+
+All of the above cases must be carefully analyzed as a whole. For example, even if the app does not permit cleartext traffic in its Android Manifest or Network Security Configuration, it might actually still be sending HTTP traffic. That could be the case if it's using a low-level API (for which Network Security Configuration is ignored) or a badly configured cross-platform framework.
```
TheDauntless, line 120:
```suggestion
All of the above cases must be carefully analyzed as a whole. For example, even if the app does not permit cleartext traffic in its Info.plist, it might actually still be sending HTTP traffic. That could be the case if it's using a low-level API (for which ATS is ignored) or a badly configured cross-platform framework.
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -104,73 +89,76 @@ Starting from January 1 2017, Apple App Store review requires justification if o
...
 
-Configuring TLS exceptions for www.example.com
+#### Testing Network API Usage
 
----
-TLSv1.3
-2019-01-15 09:39:27.892 nscurl[11459:5126999] NSURLSession/NSURLConnection HTTP load failed (kCFStreamErrorDomainSSL, -9800)
-Result : FAIL
----
+You should verify if the app allows cleartext traffic. If the source code is available, open then `Info.plist` file in the application bundle directory and look for any exceptions within the [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) optional key.
```
TheDauntless, line 136:
```suggestion
You should verify if the app allows cleartext traffic. Open the `Info.plist` file in the application bundle directory and look for any exceptions within the [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) optional key.
```
No source code is required. You can extract it from the ipa or the installed application on a jailbroken iphone.

TheDauntless: I see there's a section below that has this. Shouldn't they just be merged? Seems very unnecessary to split them up.

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -104,73 +89,76 @@ Starting from January 1 2017, Apple App Store review requires justification if o
...
+
+#### Testing Network Security Exceptions
 
-> If there are any fails in the nscurl output, please change the server side configuration of TLS to make the serverside more secure, instead of weakening the configuration in ATS on the client.
+You should verify if the app has any network security exceptions configured by opening then `Info.plist` file in the application bundle directory and inspecting the [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) optional key. Verify if it contains any exceptions and [inspect the corresponding justifications](https://developer.apple.com/documentation/security/preventing_insecure_network_connections#3138036) to discard that it might be part of the app intended purpose.
```
TheDauntless, line 154:
```suggestion
You should verify if the app has any network security exceptions configured by opening the `Info.plist` file in the application bundle directory and inspecting the [`NSAppTransportSecurity`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity) optional key. Verify if it contains any exceptions and [inspect the corresponding justifications](https://developer.apple.com/documentation/security/preventing_insecure_network_connections#3138036) to discard that it might be part of the app intended purpose.
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -104,73 +89,76 @@ Starting from January 1 2017, Apple App Store review requires justification if o
...
 
-For more information on this topic please consult the [blog post by NowSecure on ATS](https://www.nowsecure.com/blog/2017/08/31/security-analysts-guide-nsapptransportsecurity-nsallowsarbitraryloads-app-transport-security-ats-exceptions/ "A guide to ATS").
+> **Apple Recommendation: Configure Exceptions Only When Needed; Prefer Server Fixes**: It’s always better to fix the server when faced with an ATS failure. Exceptions reduce the security of your app. Some also require justification when submitting an app to the App Store, as described in the next section.
```
TheDauntless, line 156:
```suggestion
> **Apple Recommendation: Configure Exceptions Only When Needed; Prefer Server Fixes**: It’s always better to fix the server when faced with an ATS failure than to lessen the ATS configuration since exceptions reduce the security of your app. Some exceptions also require justification when submitting an app to the App Store, as described in the next section.
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -195,7 +183,30 @@ In general it can be summarized:
 ```
 
 - If connections to 3rd party domains are made (that are not under control of the app owner) it should be evaluated what ATS settings are not supported by the 3rd party domain and if they can be deactivated.
-- If the application opens third party web sites in web views, then from iOS 10 onwards `NSAllowsArbitraryLoadsInWebContent` can be used to disable ATS restrictions for the content loaded in web views
+- If the application opens third party web sites in web views, then from iOS 10 onwards `NSAllowsArbitraryLoadsInWebContent` can be used to disable ATS restrictions for the content loaded in web views.
+- If the application only need to limit ATS exception to a single domain e.g. to access the insecure server http://example.com, it can set `NSExceptionAllowsInsecureHTTPLoads` to YES in order to maintain all the benefits of ATS for all other domains.
```
TheDauntless, line 187:
```suggestion
- If the application only needs to limit ATS exceptions to a single domain e.g. to access the insecure server http://example.com, it can set `NSExceptionAllowsInsecureHTTPLoads` to YES in order to maintain all the benefits of ATS for all other domains.
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -104,73 +89,76 @@ Starting from January 1 2017, Apple App Store review requires justification if o
...
 
-In general it can be summarized:
+Consider the following when inspecting ATS:
 
 - ATS should be configured according to best practices by Apple and only be deactivated under certain circumstances.
-- If the application connects to a defined number of domains that the application owner controls, then configure the servers to support the ATS requirements and opt-in for the ATS requirements within the app. In the following example, `example.com` is owned by the application owner and ATS is enabled for that domain.
+- If the application connects to a defined number of domains that the application owner controls, then configure the servers to support the ATS requirements and opt-in for the ATS requirements within the app. In the following example, `example.com` is owned by the application owner and ATS is enabled for that domain using the [`NSExceptionDomains`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsexceptiondomains) sub-dictionary.
```
TheDauntless, line 161: Why do we talk about the NSExceptionDomains if the default is ok? I think this is legacy from older versions? Or did you write this yourself? This section is a bit confusing.

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -195,7 +183,30 @@ In general it can be summarized:
...
+- If the application only need to limit ATS exception to a single domain e.g. to access the insecure server http://example.com, it can set `NSExceptionAllowsInsecureHTTPLoads` to YES in order to maintain all the benefits of ATS for all other domains.
+
+It is possible to verify which ATS settings can be used when communicating to a certain endpoint. On macOS the command line utility `nscurl` is available to check the same. A permutation of different settings will be executed and verified against the specified endpoint. If the default ATS secure connection test is passing, ATS can be used in its default secure configuration. If there are any fails in the nscurl output, please change the server side configuration of TLS to make the serverside more secure, instead of weakening the configuration in ATS on the client. See the article "Identifying the Source of Blocked Connections" in the [Apple Developer Documentation](https://developer.apple.com/documentation/security/preventing_insecure_network_connections/identifying_the_source_of_blocked_connections) for more details.
```
TheDauntless, line 189:
```suggestion
It is possible to verify which ATS settings can be used when communicating to a certain endpoint. On macOS the command line utility `nscurl` can be used. A permutation of different settings will be executed and verified against the specified endpoint. If the default ATS secure connection test is passing, ATS can be used in its default secure configuration. If there are any fails in the nscurl output, please change the server side configuration of TLS to make the serverside more secure, rather than weakening the configuration in ATS on the client. See the article "Identifying the Source of Blocked Connections" in the [Apple Developer Documentation](https://developer.apple.com/documentation/security/preventing_insecure_network_connections/identifying_the_source_of_blocked_connections) for more details.
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -195,7 +183,30 @@ In general it can be summarized:
...
+
+For more information on ATS exceptions please consult section "Configure Exceptions Only When Needed; Prefer Server Fixes" from the article "Preventing Insecure Network Connections" in the [Apple Developer Documentation](https://developer.apple.com/documentation/security/preventing_insecure_network_connections#3138482) and the [blog post by NowSecure on ATS](https://www.nowsecure.com/blog/2017/08/31/security-analysts-guide-nsapptransportsecurity-nsallowsarbitraryloads-app-transport-security-ats-exceptions/ "A guide to ATS").
+
+### Dynamic Analysis
+
+Intercept the tested app's incoming and outgoing network traffic and make sure that this traffic is encrypted. You can intercept network traffic in any of the following ways:
+
+- Capture all HTTP(S) and Websocket traffic with an interception proxy like [OWASP ZAP](0x08-Testing-Tools.md#owasp-zap) or [Burp Suite](0x08-Testing-Tools.md#burp-suite) and make sure all requests are made via HTTPS instead of HTTP.
```
TheDauntless, line 197:
```suggestion
- Capture all HTTP(S) and Websocket traffic with an interception proxy like [OWASP ZAP](0x08-Testing-Tools.md#owasp-zap) or [Burp Suite](0x08-Testing-Tools.md#burp-suite) and make sure all requests are made via HTTPS instead of HTTP.
```
We should choose if we either always want to link Burp and Zap, or never. Or only once per testcase at the first introduction?

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -212,7 +223,7 @@ In that case, the certificate associated or 'pinned' to the host at when it seen
 
 #### When the Pin Fails
 
-Pinning failures can occur for various reasons: either the app is expecting another key or certificate than offered by the server and/or load balancer, or there might be a man-in-the-middle attack going on. In both cases and similar as with Android, there are various ways to respond to such a situation. Please see the section "[When the Pin Fails](0x05g-Testing-Network-Communication.md#when-the-pin-fails)" in the chapter "Android Network APIs".
+Pinning failures can occur for various reasons: either the app is expecting another key or certificate than offered by the server and/or load balancer, or there might be a man-in-the-middle attack going on. In both cases and similar as with Android, there are various ways to respond to such a situation. Please see the section "[When the Pin Fails](0x05g-Testing-Network-Communication.md#when-the-pin-fails)" in the chapter "Android Network Communication".
```
TheDauntless, line 226:
```suggestion
Pinning failures can occur for various reasons: either the app is expecting another key or certificate than offered by the server, or there might be a man-in-the-middle attack occuring. In both cases and similar as with Android, there are various ways to respond to such a situation. Please see the section "[When the Pin Fails](0x05g-Testing-Network-Communication.md#when-the-pin-fails)" in the chapter "Android Network Communication".
```
I think the mentioning of a loadbalancer is unnecessary here, and it would require us to mention it in many other places as well. I think it makes sense to view the external targeted system as the 'server' as a generic name.

TheDauntless (approved):
Many of them are simple accepts, a few of them are a bit more work :)

TheDauntless:
I obviously meant to click 'Request changes' rather than 'Approve' :/

Document/0x05g-Testing-Network-Communication.md
cpholguera: This was one block I simply moved around I didn't see that so thanks for telling! We have to change it. I assume it just want to say: unpack and check that value in the android manifest.

> This is one of many cases where having a "Techniques" chapter we can refer back to will come handy. This way we don't have to explain that you need to unzip the APK, decode the Android Manifest..but you can achieve the same by ...blah blah We'd simply have "Checking a value from the Android Manifest" including all alternatives. And here we'd just say: `["Check that the value minsdk is ... in the Android Manifest"](linktotechniques#checking-a-value...)`.

Document/0x06g-Testing-Network-Communication.md
cpholguera: I totally agree!

Document/0x06g-Testing-Network-Communication.md
cpholguera: I agree on that. Let's aim for consistency.

Document/0x06g-Testing-Network-Communication.md
cpholguera: I like your version more than Apple's 😁 I'll think what to do since this is verbatim. Maybe we can put some indicator of verbatim quotations.

Document/0x05b-Basic-Security_Testing.md
```diff
@@ -447,36 +460,42 @@ Note that this also shows the PID of the apps that are running at the moment. Ta
...
+- assets: directory containing app assets (files used within the Android app, such as XML files, JavaScript files, and pictures), which the [AssetManager](https://developer.android.com/reference/android/content/res/AssetManager) can retrieve
 - classes.dex: classes compiled in the DEX file format, the Dalvik virtual machine/Android Runtime can process. DEX is Java bytecode for the Dalvik Virtual Machine. It is optimized for small devices
 - lib: directory containing 3rd party libraries that are part of the APK.
 - res: directory containing resources that haven't been compiled into resources.arsc
 - resources.arsc: file containing precompiled resources, such as XML files for the layout
 
-As unzipping with the standard `unzip` utility leaves some files such as the `AndroidManifest.xml` unreadable, you better unpack the APK using apktool as described in "Recommended Tools - apktool". The unpacking results into:
+As unzipping with the standard `unzip` utility leaves some files such as the `AndroidManifest.xml` unreadable, you better unpack the APK using [apktool](0x08-Testing-Tools.md#apktool).
```
TheDauntless, line 498:
```suggestion
As unzipping with the standard `unzip` utility leaves some files such as the `AndroidManifest.xml` unreadable, it's better to unpack the APK using [apktool](0x08-Testing-Tools.md#apktool).
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -1,4 +1,130 @@
...
+## Overview
+
+### Android Network Security Configuration
+
+Starting on Android 7.0 (API level 24), Android apps can customize their network security settings using the so-called [Network Security Configuration](https://developer.android.com/training/articles/security-config) feature which offers the following key capabilities:
+
+- **Cleartext traffic**: Protect apps from accidental usage of cleartext traffic (or enables it).
+- **Custom trust anchors**: Customize which Certificate Authorities (CA) are trusted for an app's secure connections. For example, trusting particular self-signed certificates or restricting the set of public CAs that the app trusts.
+- **Certificate pinning**: Restrict an app's secure connection to particular certificates.
+- **Debug-only overrides**: Safely debug secure connections in an app without added risk to the installed base.
```
TheDauntless, line 16:
```suggestion
- **Cleartext traffic**: Protect apps from accidental usage of cleartext traffic (or enables it).
- **Custom trust anchors**: Customize which Certificate Authorities (CAs) are trusted for an app's secure connections. For example, trusting particular self-signed certificates or restricting the set of public CAs that the app trusts.
- **Certificate pinning**: Restrict an app's secure connection to particular certificates.
- **Debug-only overrides**: Safely debug secure connections in an app without added risk to the installed base.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -1,4 +1,130 @@
...
+- `domain-config` overrides `base-config` for specific domains (it can contain multiple `domain` entries).
+
+For example, the following configuration uses the `base-config` to prevent cleartext traffic for all domains. But it overrides that rule using a `domain-config`, explicitly allowing cleartext traffic for `localhost` and its subdomains.
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="false" />
+    <domain-config cleartextTrafficPermitted="true">
+        <domain includeSubdomains="true">localhost</domain>
```
TheDauntless, line 42:
```suggestion
        <domain>localhost</domain>
```
localhost can't have subdomains.

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -1,4 +1,130 @@
...
+```bash
+D/NetworkSecurityConfig: Using Network Security Config from resource network_security_config
+```
+
+The Network Security Configuration is [XML-based](https://developer.android.com/training/articles/security-config#FileFormat) and can be used to configure app-wide and domain-specific settings:
+
+- `base-config` applies to all connections that the app attempts to make.
+- `domain-config` overrides `base-config` for specific domains (it can contain multiple `domain` entries).
+
+For example, the following configuration uses the `base-config` to prevent cleartext traffic for all domains. But it overrides that rule using a `domain-config`, explicitly allowing cleartext traffic for `localhost` and its subdomains.
```
TheDauntless, line 35:
```suggestion
For example, the following configuration uses the `base-config` to prevent cleartext traffic for all domains. But it overrides that rule using a `domain-config`, explicitly allowing cleartext traffic for `localhost`.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -15,6 +141,51 @@ Make sure that the hostname and the certificate itself are verified correctly. E
 
 ### Static Analysis
 
+#### Verifying the Target SDK Version
+
+Applications targeting Android 7.0 (API level 24) or higher will use a **default Network Security Configuration that doesn't trust any user supplied CAs**, reducing the possibility of MITM attacks by luring users to install malicious CAs.
+
+[Decode the app using apktool](0x05b-Basic-Security_Testing.md#exploring-the-app-package) and verify that the `targetSdkVersion` in apktool.yml is equal or higher than `24`.
```
TheDauntless, line 148:
```suggestion
[Decode the app using apktool](0x05b-Basic-Security_Testing.md#exploring-the-app-package) and verify that the `targetSdkVersion` in apktool.yml is equal to or higher than `24`.
```

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -332,157 +480,18 @@ After decompressing the APK file, Cordova/Phonegap files will be located in the
...
+However, keep in mind that:
 
-There might still be scenarios where this is not needed and you can still do MITM attacks without patching:
+- the APIs might not be complete.
+- if nothing is hooked, that doesn't necessarily mean that the app doesn't implement pinning.
 
-- When the app is running on an Android device with Android 7.0 (API level 24) onwards, but the app targets API levels below 24, it will not use the Network Security Configuration file. Instead, the app will still trust any user supplied CAs.
-- When the app is running on an Android device with Android 7.0 (API level 24) onwards and there is no custom Network Security Configuration implemented in the app.
+In both cases, the app or some of its components might implement pinning on a very custom way that it's [not implemented in objection](https://github.com/sensepost/objection/blob/master/agent/src/android/pinning.ts).
```
TheDauntless, line 494:
```suggestion
In both cases, the app or some of its components might implement custom pinning in a way that is [supported by objection](https://github.com/sensepost/objection/blob/master/agent/src/android/pinning.ts). Please check the static analysis section for specific pinning indicators and more in-depth testing.
```

TheDauntless (changes requested):
😅

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -96,219 +104,244 @@ The following table summarizes the per-domain ATS exceptions. For more informati
...
+#### Check for Low-Level Networking API Usage
 
-In general it can be summarized:
+Identify the network APIs used by the app and see if it uses any low-level networking APIs.
 
-- ATS should be configured according to best practices by Apple and only be deactivated under certain circumstances.
-- If the application connects to a defined number of domains that the application owner controls, then configure the servers to support the ATS requirements and opt-in for the ATS requirements within the app. In the following example, `example.com` is owned by the application owner and ATS is enabled for that domain.
+> **Apple Recommendation: Prefer High-Level Frameworks in Your App**: ATS doesn’t apply to calls your app makes to lower-level networking interfaces like the Network framework or CFNetwork. In these cases, you take responsibility for ensuring the security of the connection. You can construct a secure connection this way, but mistakes are both easy to make and costly. It’s typically safest to rely on the URL Loading System instead.
```
cpholguera, line 181:
```suggestion
> **Apple Recommendation: Prefer High-Level Frameworks in Your App**: "ATS doesn’t apply to calls your app makes to lower-level networking interfaces like the Network framework or CFNetwork. In these cases, you take responsibility for ensuring the security of the connection. You can construct a secure connection this way, but mistakes are both easy to make and costly. It’s typically safest to rely on the URL Loading System instead" (see [source](https://developer.apple.com/documentation/security/preventing_insecure_network_connections)).
```

Document/0x06g-Testing-Network-Communication.md
```diff
@@ -96,219 +104,244 @@ The following table summarizes the per-domain ATS exceptions. For more informati
...
+
+- Configuring ATS to enable cleartext traffic by setting the `NSAllowsArbitraryLoads` attribute to `true` (or `YES`) on `NSAppTransportSecurity` in the app's `Info.plist`.
+- [Retrieve the `Info.plist`](0x06b-Basic-Security-Testing.md#the-infoplist-file)
+- Check that `NSAllowsArbitraryLoads` is not set to `true` globally of for any domain.
+
+- If the application opens third party web sites in web views, then from iOS 10 onwards `NSAllowsArbitraryLoadsInWebContent` can be used to disable ATS restrictions for the content loaded in web views.
+
+> **Apple warns:** Disabling ATS means that unsecured HTTP connections are allowed. HTTPS connections are also allowed, and are still subject to default server trust evaluation. However, extended security checks—like requiring a minimum Transport Layer Security (TLS) protocol version—are disabled. Without ATS, you’re also free to loosen the default server trust requirements, as described in Performing Manual Server Trust Authentication.
```
cpholguera, line 203:
```suggestion
> **Apple warns:** Disabling ATS means that unsecured HTTP connections are allowed. HTTPS connections are also allowed, and are still subject to default server trust evaluation. However, extended security checks—like requiring a minimum Transport Layer Security (TLS) protocol version—are disabled. Without ATS, you’re also free to loosen the default server trust requirements, as described in ["Performing Manual Server Trust Authentication"](https://developer.apple.com/documentation/foundation/url_loading_system/handling_an_authentication_challenge/performing_manual_server_trust_authentication).
```

PR #2055, opened 2022-02-03
Remove extra '\' after image in content

RobertDoc:
There was a random \ after the image which served no purpose.


cpholguera:
Thanks @RobertDoc, this is already being taken care of in #1979.

PR #2238, opened 2022-09-07
fix img path 

RDxR10:
N/A


cpholguera:
Hi @RDxR10, thanks for keeping an eye on potentially broken links. In this case the link is correct. Everything under docs/ is not meant to be read from GitHub. These are the sources of our website, the image is being displayed properly (see screenshot) so I'll close this PR. But please keep opening issues or PRs if you see anything else wrong. That's very helpful, thank you very much!


 

https://mas.owasp.org/news

PR #2255, opened 2022-09-27
Fix typo

CDuPlooy:
Hi, was just reading up about something and noticed the typo `Requied` instead of `Required`


cpholguera (approved):
Thank you for the fix @CDuPlooy!

Issue #2285, opened 2022-10-17
Check for Pending Intents


cpholguera:
Add theory about "Pending Intents" to 0x5h.

Add test case to check for vulnerable pending intents.

> Originally suggested in https://github.com/OWASP/owasp-mastg/pull/1962#issuecomment-956202664
> look for https://www.blackhat.com/eu-21/briefings/schedule/index.html#re-route-your-intent-for-privilege-escalation-an-universal-way-to-exploit-android-pendingintents-in-high-profile-and-system-apps-24340. They will report a new vector on how to get access to providers via PendingIntents (a lot of apps are vulnerable to that right now!)

PR #2300, opened 2022-10-27
0x05h - Android  pendingintent

### LINKED ISSUE #1874: 0x5h - Add PendingIntents
Using Intent as a mechanism to implement a callback to an Activity can lead to vulnerabilities, both in the provider and client apps. This is due to the fact that Intents are always processed in the Context of the app they are called within. This Context opens the possibility to start any non-exported activities in the provider app, and forces the client app to export the activity that should receive the callback.

PendingIntents, by contrast, are always processed inside the Context that created them. This not only allows a provider app to freely use them, without exposing non-exported activities, but also allows the client to specify any activity, including non-exported activities, to receive callbacks.

Source: https://medium.com/androiddevelopers/android-nesting-intents-e472fafc1933

https://medium.com/androiddevelopers/all-about-pendingintents-748c8eb8619

su-vikas:
Added content for PendingIntent.  

For brevity, I didn't go in details of how can a pendingintent be attacked. Let me know if it is sufficient. 

This will close #1874 and #2285


cpholguera:
Thanks @su-vikas! Nice to have you back 😃

sushi2k (approved):
LGTM

PR #2325, opened 2022-11-18
Remove deprecated method from iOS sample code

vtourraine:
Remove reference to `NSUserDefaults.synchronize` in iOS sample code, as Apple recommends not using it.

> this method is unnecessary and shouldn't be used

Source: https://developer.apple.com/documentation/foundation/nsuserdefaults/1414005-synchronize

This PR closes #2324.


cpholguera (approved):
Thank you @vtourraine!

PR #2326, opened 2022-11-22
Add iOS tcpdump instructions to 0x06b and related mitmproxy reference to 0x08a

### LINKED ISSUE #2105: Add tcpdump to 0x06b
Add tcpdump (after testing against the latest iOS) and check if there's anything else worth adding from the refs below (sslsplit?).

Refs:
- https://blog.jjhayes.net/wp/2019/02/28/capture-iphone-network-traffic-with-tcpdump-and-wireshark/
- https://iphelix.medium.com/ios-data-interception-b3a393e80c82 (2013, verify that it still works)
- https://www.mustafadur.com/blog/intercepting-ssl-and-https/ (incl. SSLSPLIT)

cgarst:
Adding content to 0x06b per #2105 which enables the reader to perform line-level packet capture for iOS devices as an alternative to RVI (suitable for non-Macs and allergies to cables).

Reviewing the references in the issue, I included most of the techniques described here. I considered SSLSPLIT, however didn't feel it would be appropriate for _this_ page of the MASTG. As I understand the scope of 0x06b to be "Basic Security Testing", I think directing those readers to modifying their Mac's pf firewall rules may be too risky for breakage - and possibly problematic with common security restrictions on work computers. Instead I recommend we consider SSLSPLIT as part of a dedicated section to include various advanced transparent interception options in a place which is agnostic to iOS/Android.

I've also added some notes regarding generating `SSLKEYLOGFILE` for HTTPS decryption in Wireshark. Since this approach requires references to mitmproxy, I have added that reference to 0x08a. If there's appetite for more mitmproxy and ZAP documentation in MASTG to reduce dependency on commercial tools that's something I'm also willing to assist with.

We may also want to open an issue to revisit the corresponding Android section in 0x05b - I believe those instructions relying on `nc` pre-date the standard `sshdump` Wireshark integration and present an approach that is a fair bit more complex and error-prone.

This PR closes #2105.


cgarst:
Hi @cpholguera any feedback on this?

cpholguera:
Hi @cgarst, I'm very sorry that this is taking so long. We're focusing on the MASVS refactoring right now but let me see if we can do a review of this PR one of these days. Thank you for your patience!

Document/0x06b-Basic-Security-Testing.md
```diff
@@ -984,8 +990,38 @@ $ rvictl -s <UDID>
...
+
+HTTPS traffic captured in the PCAP can be decrypted for review within Wireshark if testing was performed with an HTTP proxy supporting `SSLKEYLOGFILE` key logging. This is possible with [mitmproxy](https://docs.mitmproxy.org/stable/howto-wireshark-tls/), but not Burp Suite ([request](https://forum.portswigger.net/thread/option-to-create-nss-key-log-file-d57f526f)) or ZAP ([issue](https://github.com/zaproxy/zaproxy/issues/1630)). Navigate to Wireshark's `Preferences > Protocols > TLS` configuration and use the "(Pre)-Master-Secret log filename" field to browse to the `SSLKEYLOGFILE` file prepared by your HTTP proxy.
+
+Alternatively to generating `SSLKEYLOGFILE` from web proxy, the Frida script ["ios-tls-keylogger.js"](https://codeshare.frida.re/@andydavies/ios-tls-keylogger/) can sometimes be used to generate this file when iOS apps use common SSL/TLS libraries. This is considered a more fragile approach as the `CALLBACK_OFFSET` constant must be [updated in the script](https://www.mustafadur.com/blog/intercepting-ssl-and-https/#sniffing-with-sslkeylogfile) to match your tested iOS version, and not all client-side implementations of SSL/TLS will work.
```
TheDauntless, line 1020:
```suggestion
As an alternative to generating `SSLKEYLOGFILE` from the web proxy, the Frida script ["ios-tls-keylogger.js"](https://codeshare.frida.re/@andydavies/ios-tls-keylogger/) can sometimes be used to generate this file when iOS apps use common SSL/TLS libraries. This is considered a more fragile approach as the `CALLBACK_OFFSET` constant must be [updated in the script](https://www.mustafadur.com/blog/intercepting-ssl-and-https/#sniffing-with-sslkeylogfile) to match your tested iOS version, and not all client-side implementations of SSL/TLS will work.
```
Additionally, this approach works, but when/why would you need to do this? What does this ("fragile") approach offer that others don't?

TheDauntless (changes requested):
Thanks for your PR! I've got a few small comments and would love to hear your feedback :)

TheDauntless (commented):
Small updates

TheDauntless (approved):

cgarst:
Thanks for the assist @TheDauntless !

PR #2417, opened 2023-05-21
Changed value of status_cells in yaml_to_excel.py

### LINKED ISSUE #2415: OWASP_MAS_Checklist issue
Hey,

Do you plan to release newer version of OWASP MAS Checklist?
I am asking because I have a strange issue.
In the current version when I select an answer in the first point for MASVS-STORAGE-1, the color is generated few positions above the row as presented in below screen.
Does someone have similar issue?

checklist

bl13pbl03p:
The old value of status value contained `H8` which causes the issue described in #2415. See screenshot below:

Screenshot 2023-05-21 at 13 16 57

Remediation:
I changed the value of status_cells from `'H8:L400'` to `'I13:L400'`.

Effectiveness: 
I am not entirely sure the if the commit will help, was not able to generate a .xlsx checklist locally.

References:
The only reference that I have used was an earlier commit 3bfc72ebb60bfd6b09051568481a3343781b68a1, looking at the value of status_cells.

Possible next steps:
A separate rule should be made to fix 'status dropdown bar issue' in the remaining tabs (MASVS-AUTH MASVS-CODE), because status dropdown function is not applicable there on I13.

This PR closes #2415 except maybe for the MASVS-AUTH and MASVS-CODE tab.

### DIFF
--- a/tools/scripts/yaml_to_excel.py
+++ b/tools/scripts/yaml_to_excel.py
@@ -153,7 +153,7 @@ def create_security_requirements_sheet(wb):
         write_header(ws, category_title, mas_styles.MASVS_COLORS[group_id])
         set_columns_width(ws)
 
-        status_cells = 'H8:L400'
+        status_cells = 'I13:L400'
         ws.conditional_formatting.add(status_cells, mas_styles.rule_fail)
         ws.conditional_formatting.add(status_cells, mas_styles.rule_pass)
         ws.conditional_formatting.add(status_cells, mas_styles.rule_na)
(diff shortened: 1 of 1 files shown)


cpholguera:
Thanks a lot for opening the PR, @bl13pbl03p! I'll take a look this week.

cpholguera (approved):
@bl13pbl03p thank you so much for your help. It should be fixed now!

tools/scripts/yaml_to_excel.py
```diff
@@ -153,7 +153,7 @@ def create_security_requirements_sheet(wb):
         write_header(ws, category_title, mas_styles.MASVS_COLORS[group_id])
         set_columns_width(ws)
 
-        status_cells = 'H8:L400'
+        status_cells = 'I13:I400'
```
cpholguera, line 156: Another fix so that we can also mark the first control, not only tests.
```suggestion
        status_cells = 'I11:I400'
```

PR #2427, opened 2023-07-29
Proofreading fixes 0x05d part 1 

cpholguera:


0xroot-bf (approved):
Narrative looks good to me, also checked that URLs were still working

PR #2478, opened 2024-01-09
Fix modified dates in docs

philyoosays:
## Description
This PR adds a fetch to the [Github REST API](https://docs.github.com/en/rest?apiVersion=2022-11-28) to collect the last modified date for [commits](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits) to MASTG markdown files for display in the [MASTG mkdocs site](https://mas.owasp.org/MASTG/tests/).

As a fallback in case the `GITHUB_TOKEN` was not able to be retrieved from the Github Actions pipeline, the original code to get last modified dates was preserved and called.

## Issue

The issue being fixed has to do with the last modified dates being incorrect in the published [mkdocs site](https://mas.owasp.org/MASTG/tests/). It seems that the Github Actions pipeline is copying the files and the original code that determines the last modified dates is using the local terminal `git log` method to determine last modified date. This is causing all the files to have the same last modified date (date of merge to master for any file) that does not match the actual last modified dates for the files in the repository.

## The Fix

To be sure that we have an accurate commit history, this fix uses the Github REST API to fetch the commit history for each markdown file from the repository.

**Note To Maintainers**: Please make sure that the `GITHUB_TOKEN` has read access to [content](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

---

Thank you for submitting a Pull Request to the OWASP MASTG. Please make sure that:

This PR closes #< insert number here >.

### DIFF
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -21,9 +21,7 @@ jobs:
 
       - run: ./src/scripts/structure_mastg.sh
 
-      - env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: python3 src/scripts/transform_files.py
+      - run: python3 src/scripts/transform_files.py
 
       - name: Get Latest MASVS Release Tag
         run: echo "MASVS_VERSION=$(curl -s https://api.github.com/repos/OWASP/owasp-masvs/releases/latest | jq '.tag_name' | sed 's/\"//g')" >> $GITHUB_ENV
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -21,7 +21,9 @@ jobs:
 
       - run: ./src/scripts/structure_mastg.sh
 
-      - run: python3 src/scripts/transform_files.py
+      - env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: python3 src/scripts/transform_files.py
 
       - name: Get Latest MASVS Release Tag
         run: echo "MASVS_VERSION=$(curl -s https://api.github.com/repos/OWASP/owasp-masvs/releases/latest | jq '.tag_name' | sed 's/\"//g')" >> $GITHUB_ENV
--- a/src/scripts/git_data.py
+++ b/src/scripts/git_data.py
@@ -4,19 +4,6 @@ import os
 import requests
 import subprocess
 
-def get_last_commit_date_by_terminal_log(file_path):
-    try:
-        # get the last commit date as "September 12, 2022"
-        command = f"git log -n 1 --date=format:'%B %d, %Y' --format=%ad -- {file_path}"
-        result = subprocess.check_output(command, shell=True, universal_newlines=True)
-
-        return result.strip()
-    except subprocess.CalledProcessError as e:
-        print(f"Error executing Git command: {e}")
-    except Exception as e:
-        print(f"An error occurred: {e}")
-    return None
-
 def get_last_commit_date(file_path):
     try:
         # get the last commit date as "September 12, 2022"
@@ -24,7 +11,7 @@ def get_last_commit_date(file_path):
 
         if GITHUB_TOKEN is None:
             print('Github token not found')
-            return get_last_commit_date_by_terminal_log(file_path)
+            return None
         
         headers = {
             'X-GitHub-Api-Version': '2022-11-28',
--- a/src/scripts/git_data.py
+++ b/src/scripts/git_data.py
@@ -1,3 +1,7 @@
+from datetime import datetime
+import json
+import os
+import requests
 import subprocess
 
 def get_last_commit_date_by_terminal_log(file_path):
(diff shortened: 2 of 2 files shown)


src/scripts/git_data.py
```diff
@@ -13,6 +17,36 @@ def get_last_commit_date(file_path):
...
+            print('Github token not found')
+            return get_last_commit_date_by_terminal_log(file_path)
+        
+        headers = {
+            'X-GitHub-Api-Version': '2022-11-28',
+            'Accept': 'application/vnd.github+json',
+            'Authorization': f'Bearer {GITHUB_TOKEN}'
+        }
+
+        url = f'https://api.github.com/repos/OWASP/owasp-mastg/commits?path={file_path}'
```
cpholguera, line 35: It was not working on local without this. It tried to request the local path to the API.
```suggestion
        path = file_path.split('owasp-mastg/')[1]
        url = f'https://api.github.com/repos/OWASP/owasp-mastg/commits?path={path}'
```

src/scripts/git_data.py
```diff
@@ -13,6 +17,36 @@ def get_last_commit_date(file_path):
         print(f"An error occurred: {e}")
     return None
 
+def get_last_commit_date(file_path):
+    try:
+        # get the last commit date as "September 12, 2022"
+        GITHUB_TOKEN = os.environ['GITHUB_TOKEN']
```
cpholguera, line 23: This way it has the chance of returning None. Otherwise it raises an exception.
```suggestion
        GITHUB_TOKEN = os.getenv('GITHUB_TOKEN')
```

src/scripts/git_data.py
```diff
@@ -13,6 +17,36 @@ def get_last_commit_date(file_path):
...
+
+        if GITHUB_TOKEN is None:
+            print('Github token not found')
+            return get_last_commit_date_by_terminal_log(file_path)
+        
+        headers = {
+            'X-GitHub-Api-Version': '2022-11-28',
```
cpholguera, line 30: is this really needed?

src/scripts/git_data.py
```diff
@@ -13,6 +17,36 @@ def get_last_commit_date(file_path):
...
     return None
 
+def get_last_commit_date(file_path):
+    try:
+        # get the last commit date as "September 12, 2022"
+        GITHUB_TOKEN = os.environ['GITHUB_TOKEN']
+
+        if GITHUB_TOKEN is None:
+            print('Github token not found')
+            return get_last_commit_date_by_terminal_log(file_path)
```
cpholguera, line 27: we can get rid of that function and return None or "N/A" here.

src/scripts/git_data.py
```diff
@@ -13,6 +17,36 @@ def get_last_commit_date(file_path):
...
+        
+        headers = {
+            'X-GitHub-Api-Version': '2022-11-28',
+            'Accept': 'application/vnd.github+json',
+            'Authorization': f'Bearer {GITHUB_TOKEN}'
+        }
+
+        url = f'https://api.github.com/repos/OWASP/owasp-mastg/commits?path={file_path}'
+        response = requests.get(url, headers=headers)
+        data = json.loads(response.text)
```
cpholguera, line 37: Better to handle the status code. Also the `.json()` method from requests is more efficient.

```suggestion
        if response.status_code != 200:
            response.raise_for_status()

        data = response.json()
```

src/scripts/git_data.py
```diff
@@ -13,6 +17,36 @@ def get_last_commit_date(file_path):
...
+
+        if 'message' in data:
+            raise Exception(data['message'])
+        elif len(data) and data[0]['commit']:
+            commit_date_in_utc = data[0]['commit']['committer']['date']
+            fmt_date = datetime.strptime(commit_date_in_utc, '%Y-%m-%dT%H:%M:%SZ').strftime('%B %d %Y')
```
cpholguera, line 43:
```suggestion
            fmt_date = datetime.strptime(commit_date_in_utc, '%Y-%m-%dT%H:%M:%SZ').strftime('%B %d, %Y')
```

.github/workflows/build-website.yml
```diff
@@ -21,7 +21,9 @@ jobs:
 
       - run: ./src/scripts/structure_mastg.sh
 
-      - run: python3 src/scripts/transform_files.py
+      - env:
```
cpholguera, line 24: It might work without this. The GITHUB_TOKEN is created automatically on each run and should be accessible to the script via env.

cpholguera (changes requested):
Thanks @philyoosays! Here a couple of suggestions. Apart from that I'm just worried about the time that it takes to retrieve all data for all files (which can be hundreds). Let's discuss on Slack.

src/scripts/git_data.py
philyoosays: The call works without it but the documentation had specified it. https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api?apiVersion=2022-11-28#4-make-a-request-with-github-cli

Would you like to remove it?

.github/workflows/build-website.yml
philyoosays: Let's try. We can always revert the commit that removes this declaration

src/scripts/git_data.py
philyoosays: ae35db4731c468ac1222cdb723d49afb174fad08

.github/workflows/build-website.yml
philyoosays: fa400261a285969b7777d533403d34db6425b17d

src/scripts/git_data.py
cpholguera: That's fine, thanks for clarifying!

cpholguera:
Fixed with depth=0 in https://github.com/OWASP/owasp-mastg/pull/2564

PR #2564, opened 2024-02-20
Set depth to 0 & add last updated to the tests table

cpholguera:


anantshri:
@cpholguera curious why are we going for fetch-depth = 0 

https://github.com/actions/checkout?tab=readme-ov-file#Fetch-all-history-for-all-tags-and-branches

depth zero is only recommended if we really need all of history. depth = 1 or others are use to reduce the disk and processing time if i recall right?

cpholguera:
Hi @anantshri, that's correct. However, this is the only way we have come up with to display "Last Updated" as you can see here: https://mas.owasp.org/MASTG/tests/ 

If you have other ideas to achieve the same, they are very welcome!

anantshri:
ah we do need history in that case unless we maintain a file somewhere in the repository that gets this details updated in it when an entry change but i guess that might be a bit of overkill. 

thanks for taking time to update here.

cpholguera:
@anantshri, thank _you_ for keeping an eye on the project. I really appreciate it. Please don't hesitate to report such things in the future.

PR #2572, opened 2024-03-01
Updated L1 vs L2 recommentation for MASVS-STORAGE

cpholguera:
The MASTG in its version 1.7.0 and earlier recommended to encrypt all sensitive data in the internal memory of an app, even for L1 apps. However, the internal memory on Android and iOS is considered safe enough. Therefore, we are updating the recommendation to reflect this and allow L1 apps to store unencrypted data in internal storage. L2 apps must also encrypt data on internal storage. The recommendation for external storage does not change.

### DIFF
--- a/tests/android/MASVS-STORAGE/MASTG-TEST-0001.md
+++ b/tests/android/MASVS-STORAGE/MASTG-TEST-0001.md
@@ -18,9 +18,9 @@ This test case focuses on identifying potentially sensitive data stored by an ap
 - Analyze data storage in the source code.
 - Be sure to trigger all possible functionality in the application (e.g. by clicking everywhere possible) in order to ensure data generation.
 - Check all application generated and modified files and ensure that the storage method is sufficiently secure.
-    - This includes SharedPreferences, SQL databases, Realm Databases, Internal Storage, External Storage, etc.
-  
-In general sensitive data stored locally on the device should always be at least encrypted, and any keys used for encryption methods should be securely stored within the Android Keystore. These files should also be stored within the application sandbox. If achievable for the application, sensitive data should be stored off device or, even better, not stored at all.
+    - This includes `SharedPreferences`, databases, Internal Storage, External Storage, etc.
+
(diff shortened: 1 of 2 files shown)


sushi2k (approved):

gagnonca:
Hey @cpholguera, I agree with the new wording on MASTG-TEST-0052 and am wondering if the language should be updated for to include language about [UIDocumentPickerViewController](https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller). 

To my knowledge, this is the only way of storing data outside of the sandbox in iOS in a way that arbitrary apps would be able to also read. The other app would also need to implement `UIDocumentPickerViewController` and the user would need to manually select the file so less of an issue than android external storage. Happy to do a PR myself if you agree.

cpholguera:
Hey @gagnonca, that's a great point, thanks for bringing it up!

As we move towards MASTG v2, we'll be creating new "weaknesses" and porting the existing v1 tests into v2 tests. This means that we won't be maintaining the current tests and will be offering a new version of it that is vastly improved and more comprehensive in comparison.


This is actually a perfect timing from your side. I'm currently preparing a document to help us identify and scope the weaknesses we are addressing when dealing with an application using the `UIDocumentPickerViewController` and other APIs.

Do you mind if I share it with you (possibly next week or over the next week) to get your feedback? Once we've defined these weaknesses, it would be great if you wanted to contribute with a PR as well. 

Meanwhile, in case you missed it, here's the link to the talk and the new MASWE.

Video: https://drive.google.com/file/d/1Fb9A4RbIVYetmgstrJV_x7MfPaJJJHBP/view?usp=sharing

Slides: https://drive.google.com/file/d/1IFfrJQjIOHkFG4UR95VIgtVPzi7wvvJ7/view?usp=drive_link

MASWE: https://mas.owasp.org/MASWE/

gagnonca:
@cpholguera sounds great! I won't get a chance to look or do anything until next week. I reference this test for iOS from ASA storage requirement here (2.1.1.1):

https://github.com/appdefensealliance/ASA-WG/blob/main/Mobile%20App%20Profile/Mobile%20App%20Test%20Guide.md

cpholguera:
That's not a problem, thanks for the context. I'll take a look and I'll see if I can share the doc with you by next week to continue the discussion there. Thanks @gagnonca!

PR #2661, opened 2024-07-08
Fix local build pipeline

TheDauntless:
* adds -p flags to avoid errors
* Duplicate gsed check to structure_masvs.sh
* Change `mv` to `cp` to leave owasp-masvs intact

This should be tested on the GitHub build workflow before merging.


cpholguera (approved):
Thank you!!

PR #2662, opened 2024-07-08
[TOOL] ipsw

TheDauntless:
I was looking through the style guide for code blocks, and I'm not entirely sure how we should document the generic tool commands. The guide says to use placeholders for arguments, but then we can't show any useful output. Or we _only_ use placeholders, but the output seems useful in this case.

I also played around with the fold-out admonition for the full command output, though I didn't add it for swift-dump since it would be 18k lines of code.

Finally, I added a 'host' meta tag. Could be useful info, no?

### DIFF
--- a/tools/ios/MASTG-TOOL-0103.md
+++ b/tools/ios/MASTG-TOOL-0103.md
@@ -12,7 +12,7 @@ IPSW calls itself an "iOS/macOS Research Swiss Army Knife". In general, IPSW all
 IPSW can extract Objective-C class information from a MachO binary. The desired architecture can be specified using `--arch` in case of a universal MachO file:
 
 ```bash
-ipsw class-dump --arch arm64 UnCrackable\ Level\ 1
+$ ipsw class-dump --arch arm64 UnCrackable\ Level\ 1
 
 @protocol NSObject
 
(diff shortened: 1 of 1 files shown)


tools/ios/MASTG-TOOL-0103.md
```diff
@@ -0,0 +1,374 @@
+---
+title: IPSW
+platform: ios
+source: https://github.com/blacktop/ipsw
+host: [windows, linux, macOS]
+---
+
+IPSW calls itself an "iOS/macOS Research Swiss Army Knife". In general, IPSW allows you to obtain iOS specific files (IPSW, OTA, ...) and also statically analyze them. For application analysis, the most interesting features are the Objective-C and Swift class-dumps. Other features are available (kernelcache parser, device-tree parser, disassembler, etc) but are only useful if you're analyzing the security of the OS itself.
+
+## Extracting Objective-C Class Information
```
cpholguera, line 10: These kind of things belong in the corresponding MASTG-TECH. This way once we enable xrefs for these components we'll be able to navigate to these examples of use right from the tool.

cpholguera (changes requested):

cpholguera:
added in https://github.com/OWASP/owasp-mastg/pull/2849

PR #2803, opened 2024-07-13
Improve structuring scripts and flatten nav

cpholguera:
By flattening the nav for MASTG and MASVS we can simplify the structuring scripts.

Also further optimized them e.g. using loops in certain cases and less copying.


TheDauntless:
Works for me locally. I did have to merge these changes with the new github workflow changes, so we need to check the build pipline after merging this.

TheDauntless (approved):

PR #2827, opened 2024-07-16
[MASTG-TOOL-0100] Update tool name to reFlutter

cpholguera:
The tool is called reFlutter and not re-flutter.


TheDauntless:
That explains it :D

TheDauntless (approved):

PR #3035, opened 2024-11-05
Port MASTG-TEST-0022: Testing Custom Certificate Stores and Certificate Pinning (android) (by @guardsquare)

### LINKED ISSUE #2958: MASTG v1->v2 MASTG-TEST-0022: Testing Custom Certificate Stores and Certificate Pinning (android)

- **Title:** Testing Custom Certificate Stores and Certificate Pinning
- **ID:** MASTG-TEST-0022
- **Link:** https://mas.owasp.org/MASTG/tests/android/MASVS-NETWORK/MASTG-TEST-0022/
- **Platform:** android
- **MASVS v1:** ['MSTG-NETWORK-4']
- **MASVS v2:** ['MASVS-NETWORK-2']

titze:
This PR closes #2958.

Couple of notes:
- I did not move the original dynamic analysis, as I believe this does not really test if pinning is performed, or would give much information (e.g., for which domains). happy to discuss
- The old test also had in its headline "Testing Custom Certificate Stores". This is not really there anymore now. The Network Security Configuration does not allow this, and I am not sure there is even a good reason to do this at all. The MASWE-0047 is also about pinning only


tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-1.md
```diff
@@ -0,0 +1,38 @@
...
+
+Apps can configure certificate pinning using the [Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning"). For each domain, one or multiple digests can be pinned.
+
+Certificate pinning can also be done manually in the code. Depending on the used technologies, this can be done for example by:
+
+- Pinning a certificate with a custom `TrustManager`,
+- configuring the used third party networking libraries to pin certificates,
+- use plugins to achieve certificate pinning for hybrid apps.
+
+Chapter [Certificate pinning without Android Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning-without-android-network-security-configuration") explains in more detail how this can be achieved in the app.
```
sushi2k, line 19:
```suggestion
```

If we remove it in the document, let's also remove the reference here.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-1.md
```diff
@@ -0,0 +1,38 @@
...
+
+The goal of this test is to check if any certificate pinning exists.
```
sushi2k, line 21:
```suggestion
The goal of this test is to verify if any certificate pinning implementation is present in the code.
```

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-1.md
```diff
@@ -0,0 +1,38 @@
...
+weakness: MASWE-0047
+---
+
+## Overview
+
+Apps can configure certificate pinning using the [Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning"). For each domain, one or multiple digests can be pinned.
+
+Certificate pinning can also be done manually in the code. Depending on the used technologies, this can be done for example by:
+
+- Pinning a certificate with a custom `TrustManager`,
```
sushi2k, line 15:
```suggestion
- Pinning a certificate with a custom `TrustManager` (which should be avoided and the NSC should be used instead),
```

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-1.md
```diff
@@ -0,0 +1,38 @@
...
+- configuring the used third party networking libraries to pin certificates,
+- use plugins to achieve certificate pinning for hybrid apps.
+
+Chapter [Certificate pinning without Android Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning-without-android-network-security-configuration") explains in more detail how this can be achieved in the app.
+
+The goal of this test is to check if any certificate pinning exists.
+
+!!! note "Limitations"
+    Since there are many different ways to achieve certificate pinning in the code, checking statically if the application performs pinning might not reveal all such locations. To make sure certificates are pinned for all relevant connections, additional dynamic analysis can be performed.
```
sushi2k, line 24:
```suggestion
    Since there are many different ways to achieve certificate pinning in the code, checking statically if the application performs pinning might not reveal all such locations. To make sure certificates are pinned for all relevant connections, additional dynamic analysis need to be performed.
```

sushi2k: Dynamic analysis is a must to effectively test Certificate Pinning.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-2.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Expired Certificate Pins
+platform: android
+id: MASTG-TEST-0x22-2
```
sushi2k, line 4: Could you please update each test and give it a unique ID like here: https://github.com/OWASP/owasp-mastg/tree/master/tests-beta/android/MASVS-NETWORK. This means also changing the filename accordingly.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-3.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Missing Certificate Pinning in Network Traffic 
+platform: network
+id: MASTG-TEST-0x22-3
```
sushi2k, line 4: Could you please update each test and give it a unique ID like here: https://github.com/OWASP/owasp-mastg/tree/master/tests-beta/android/MASVS-NETWORK. This means also changing the filename accordingly.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-2.md
```diff
@@ -0,0 +1,26 @@
...
+type: [static]
+weakness: MASWE-0047
+---
+
+## Overview
+
+Apps can configure expiration dates for pinned certificates in the [Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning"). After the expiration date the pin is not used any more, and all installed CAs are trusted for that domain.
```
sushi2k, line 11:
```suggestion
Apps can configure expiration dates for pinned certificates in the [Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning"). After the expiration date, the pin is no longer used and all installed CAs for that domain are trusted.
```

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-3.md
```diff
@@ -0,0 +1,29 @@
...
+type: [static]
+weakness: MASWE-0047
+---
+
+## Overview
+
+There are various ways how [certificate pinning]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning-without-android-network-security-configuration") can be done for an application.
```
sushi2k, line 11:
```suggestion
```

If we remove it in the document, let's also remove the reference here.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-3.md
```diff
@@ -0,0 +1,29 @@
...
+
+Since statically finding all of the locations where certificate pinning is performed might not be feasible, this test case uses dynamic analysis to observe all connections the app makes.
+
+The goal of this test case is to dynamically check if the connection to a server can be intercepted using a [Man-in-the-Middle attack]("../../../Document/0x04f-Testing-Network-Communication.md#mitm-attack). If this is possible, it means the certificate is not pinned correctly or at all.
```
sushi2k, line 15:
```suggestion
The goal of this test case is to dynamically check if the connection to a server can be intercepted using a [Machine-in-the-Middle attack]("../../../Document/0x04f-Testing-Network-Communication.md#mitm-attack). If this is possible, it means that the certificate is not pinned correctly or not pinned at all.
```

sushi2k (changes requested):
Thank you for the new test cases and the PR @titze! Please check the suggestions below.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0x22-3.md
titze: I think the text itself does make sense, but if the link is gone, it makes sense to remove that.

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -119,6 +119,36 @@ If at least one of the pinned digests matches, the certificate chain will be con
...
+```java
+val client = OkHttpClient.Builder()
+      .certificatePinner(
+          CertificatePinner.Builder()
+              .add("publicobject.com", "sha256/afwiKY3RxoMmLkuRW1l7QsPZTJPwDS2pdDROQjXw8ig=")
+              .build())
+      .build()
+```
+
+#### Hybrid Applications
```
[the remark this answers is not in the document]
titze: Makes sense.

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -119,6 +119,36 @@ If at least one of the pinned digests matches, the certificate chain will be con
 </network-security-config>
 ```
 
+!!! note "Expiration dates"
+    If you [set an expiration date](https://developer.android.com/privacy-and-security/security-config#CertificatePinning), make sure to update your application in time. Otherwise pinning will **not** be performed at all after the configured date.
+
+!!! warning "Technologies not using Network Security Configuration"
+    If your application uses low level networking APIs or SDKs like Flutter, the Network Security Configuration might not be used by default. In those cases you need to enable certificate pinning specifically for the used technology.
+
+### Certificate pinning without Android Network Security Configuration
```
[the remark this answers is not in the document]
titze: For Android 7.0 I agree.

But what about other communication libraries like `okhttp`? I think it would be valuable to read about those, and that you can configure this differently for them, wdyt?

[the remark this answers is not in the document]
cpholguera: I agree we should keep the mention. Maybe this wasn't clear due to how the section is currently structured: it's a subsection of "### Android Network Security Configuration"

I think we should have a dedicated section: "### Certificate Pinning"

And have subsections for each method.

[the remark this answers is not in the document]
cpholguera: I asked perplexity to draft one for us, it's nice because it also includes one method we don't have here yet: TrustManager, and more.

@titze could you use this as reference to rework this section? (Carefully review and rewrite as needed since this is fully AI generated; the markdown header levels also don't match). Recommendations shouldn't be here, consider reviewing the corresponding MASWE and creating a new MASTG-BEST-xxxx to capture them and link from the tests.


--------


# Certificate Pinning  

Certificate pinning is a critical security mechanism employed in Android applications to safeguard against man-in-the-middle (MITM) attacks by ensuring that the app communicates exclusively with servers possessing predefined cryptographic credentials[1][3]. While effective when implemented correctly, variations in implementation methodologies can introduce vulnerabilities, particularly when developers misuse platform-specific APIs or third-party libraries[2][4]. Below is an analysis of common implementation approaches, their security implications, and recommendations for robust deployment.  

## Implementation Methods for Certificate Pinning in Android  

### Network Security Configuration (API 24+)  
Introduced in Android Nougat (API 24), the network security configuration file (`network_security_config.xml`) provides a declarative method for certificate pinning without requiring extensive custom code[3]. Developers specify public key hashes (pins) for trusted domains directly in XML, allowing the Android operating system to validate server certificates during the TLS handshake. This method is inherently secure because it leverages the OS’s certificate validation routines, ensuring that only the cleaned certificate chain—free of attacker-injected certificates—is evaluated against the pins[3]. For example, pins derived from a server’s public key remain valid even if the certificate is reissued with the same key, reducing maintenance overhead[3].  

### OkHttp Library  
The OkHttp library is widely used for certificate pinning in Android due to its built-in `CertificatePinner` class, which enables developers to pin public key hashes or certificate signatures[1][3]. When configured correctly—by pinning to intermediate or root certificates rather than leaf certificates—OkHttp provides robust protection against MITM attacks. However, insecure implementations arise when developers hardcode leaf certificate hashes, which become invalid upon certificate rotation, necessitating frequent app updates[3]. Additionally, older versions of OkHttp were vulnerable to chain-cleaning flaws if they relied on `getPeerCertificates()`, which returns server-sent certificates rather than the validated chain[2]. Upgrading to the latest OkHttp version and pinning to public keys mitigates these risks[2].  

### TrustKit for Backward Compatibility  
For Android versions prior to API 24, the TrustKit library offers a backward-compatible solution by integrating certificate pinning into the app’s network layer[3]. TrustKit validates pins against the cleaned certificate chain and supports report-uri functionality to notify developers of pin validation failures. However, improper configuration—such as enabling non-default options like CA pinning—can reintroduce vulnerabilities akin to custom implementations[2]. For instance, the PhoneGap SSL Certificate Checker plugin’s CA-pinning feature was found to be bypassable until patched[2].  

### Custom Implementations Using HttpsURLConnection  
Developers occasionally implement certificate pinning manually by overriding `TrustManager` or `HostnameVerifier` in `HttpsURLConnection`[1][2]. This approach is highly error-prone, as it often involves directly inspecting server-sent certificates via `getPeerCertificates()`, which returns unvalidated chains that attackers can manipulate[2]. For example, an attacker could present a malicious certificate chain containing a trusted root CA, causing the app to mistakenly validate a rogue leaf certificate[2]. Secure custom implementations require reconstructing the OS-validated certificate chain, a complex process prone to implementation errors[2].  

### Vulnerable Third-Party Libraries  
Third-party libraries like older versions of the Secure-HTTP Cordova plugin or misconfigured PhoneGap plugins historically introduced vulnerabilities by mishandling certificate chains[2]. These libraries often exposed non-default configuration options that, when enabled, allowed attackers to bypass pinning by injecting untrusted certificates into the chain. Developers must audit third-party libraries for adherence to chain-validation best practices and prefer those leveraging the Android Keystore system for pinning[2].  

## Recommendations for Secure Certificate Pinning

1. **Prefer Network Security Configuration**: For apps targeting API 24+, use the built-in `network_security_config.xml` to delegate pinning to the OS, minimizing human error[3].  
2. **Update Libraries Regularly**: Ensure OkHttp, TrustKit, and other third-party libraries are updated to versions that validate cleaned certificate chains[2][3].  
3. **Avoid Leaf Certificate Pinning**: Pin to public keys of intermediate or root CAs to accommodate certificate rotations without app updates[3].  
4. **Deprecate Custom Implementations**: Refactor legacy code using `HttpsURLConnection` or `TrustManager` to use vetted libraries or the network security config[2].  
5. **Monitor for Pinning Failures**: Implement report-uri endpoints or logging mechanisms to detect and respond to pin validation failures, indicative of ongoing attacks[3].  

By adhering to these practices, developers can mitigate risks associated with outdated or insecure pinning implementations while maintaining compatibility across Android versions[4].

## Citations
[1] What is Certificate Pinning? - Zimperium https://www.zimperium.com/glossary/certificate-pinning/
[2] Ineffective Certificate Pinning Implementations | Black Duck Blog https://www.blackduck.com/blog/ineffective-certificate-pinning-implementations.html
[3] Blog: Securing HTTPS with Certificate Pinning on Android - Approov https://approov.io/blog/securing-https-with-certificate-pinning-on-android
[4] modern alternatives to outdated certificate pinning practices https://blog.cloudflare.com/why-certificate-pinning-is-outdated/
[5] Mobile Development and Security: Certificate Pinning on Android https://nglogic.com/cert-pinning-android/
[6] 3 Ways How To Implement Certificate Pinning on Android - Netguru https://www.netguru.com/blog/android-certificate-pinning
[7] Certificate Pinning — Does It Help App Security? – PreEmptive https://www.preemptive.com/blog/what-is-certificate-pinning-app-security/
[8] Enhancing Security in Android Applications with SSL Pinning https://www.droidcon.com/2024/07/11/enhancing-security-in-android-applications-with-ssl-pinning/
[9] Defeating Android Certificate Pinning with Frida - HTTP Toolkit https://httptoolkit.com/blog/frida-certificate-pinning/
[10] Higher risk of no certificate pinning on mobile apps vs web apps? https://security.stackexchange.com/questions/226062/higher-risk-of-no-certificate-pinning-on-mobile-apps-vs-web-apps
[11] Android SSL certificate pinning to prevent Man-in-the-middle attack https://blog.ostorlab.co/android-ssl-pinning.html
[12] What is the best way to implement SSL Pinning? : r/androiddev https://www.reddit.com/r/androiddev/comments/uuie96/what_is_the_best_way_to_implement_ssl_pinning/
[13] Certificate Pinning - What it Is, its Benefits, and Drawbacks https://www.sapphire.net/blogs-press-releases/certificate-pinning/
[14] TLS Certificate Security for Android | Guardsquare https://www.guardsquare.com/blog/insecure-tls-certificate-checking-in-android-apps
[15] The problem with certificate pinning | SSLs.com Blog https://www.ssls.com/blog/the-problem-with-certificate-pinning/
[16] Security with network protocols - Android Developers https://developer.android.com/privacy-and-security/security-ssl
[17] Certificate Pinning: Challenges and Viable Alternatives - SecureW2 https://www.securew2.com/blog/everything-you-need-to-know-about-certificate-pinning
[18] Unpacking Android Security: Part 3 — Insecure Communication https://proandroiddev.com/unpacking-android-security-part-3-insecure-communication-d5e833166040
[19] Security with network protocols - Android Developers https://developer.android.com/privacy-and-security/security-ssl
[20] Certificate Pinning, Public-Key Pinning oder CA Pinning? https://www.cologne-intelligence.de/blog/pinning-in-android-apps
[21] Secure Android Apps with TLS/SSL Pinning - ProAndroidDev https://proandroiddev.com/secure-android-apps-with-tls-ssl-pinning-c087fc7ef828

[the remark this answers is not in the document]
titze: I added / merged most of this into the existing doc, and cleaned it up a bit.
I did not add any recommendations, since those are already captured in https://mas.owasp.org/MASWE/MASVS-NETWORK/MASWE-0047/#mitigations.

cpholguera (changes requested):
Hi @titze there's a problem with the PR, it now affects 240 files. Could you please recheck/fix the merging of the main branch so that only your changes are showing?

titze:
Fixed. I messed up the rebase on master...

cpholguera:
No problem, thanks so much for the fix!

cpholguera (changes requested):

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -119,6 +127,48 @@ If at least one of the pinned digests matches, the certificate chain will be con
...
+
+!!! warning "Implementing Certificate Pinning Manually"
+    Implementing certificate pinning manually has a high risk of adding functionality to your application that makes the app even less secure. If you are adding this manually, take extreme care of implementing this correctly.
+
+### Vulnerable Third-Party Libraries
+
+Third-party libraries like older versions of the Secure-HTTP Cordova plugin or misconfigured PhoneGap plugins historically introduced vulnerabilities by [mishandling certificate chains](https://www.blackduck.com/blog/ineffective-certificate-pinning-implementations.html). These libraries often exposed non-default configuration options that, when enabled, allowed attackers to bypass pinning by injecting untrusted certificates into the chain.
+
+Developers must audit third-party libraries for adherence to chain-validation best practices and prefer those leveraging the Android Keystore system for pinning.
```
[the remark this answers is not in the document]
cpholguera: **TODO**: Add a short section about "Pinning in Cross-Platform Frameworks" e.g. Flutter, Xamarin, React Native, Cordova/Ionic (at least some considerations). For example Flutter supports NSC, see https://docs.flutter.dev/release/breaking-changes/network-policy-ios-android#migration-guide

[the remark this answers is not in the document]
cpholguera: **TODO**: Add a short section about "Pinning in Cross-Platform Frameworks" e.g. Flutter, Xamarin, React Native, Cordova/Ionic (at least some considerations). For example Flutter supports NSC, see https://docs.flutter.dev/release/breaking-changes/network-policy-ios-android#migration-guide

sushi2k (commented):
One small change from my side. Otherwise this looks good to me. 

Thank you for the work @titze  and the suggestions @cpholguera!

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
...
+weakness: MASWE-0047
+---
+
+## Overview
+
+There are various ways how certificate pinning can be done for an application.
+
+Since statically finding all of the locations where certificate pinning is performed might not be feasible, this test case uses dynamic analysis to observe all connections the app makes.
+
+The goal of this test case is to dynamically check if the connection to a server can be intercepted using a [Man-in-the-Middle attack]("../../../Document/0x04f-Testing-Network-Communication.md#mitm-attack). If this is possible, it means that the certificate is not pinned correctly or not pinned at all.
```
sushi2k, line 15:
```suggestion
The goal of this test case is to dynamically check if the connection to a server can be intercepted using a [Machine-in-the-Middle attack]("../../../Document/0x04f-Testing-Network-Communication.md#mitm-attack). If this is possible, it means that the certificate is not pinned correctly or not pinned at all.
```

sushi2k (commented):
MiTM change

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -119,6 +127,48 @@ If at least one of the pinned digests matches, the certificate chain will be con
...
+      .build()
+```
+
+### Certificate Pinning using TrustKit
+
+For Android versions prior to API 24, the [TrustKit library](https://github.com/datatheorem/TrustKit-Android) offers a backward-compatible solution. TrustKit validates pins against the cleaned certificate chain. However, improper configuration — such as enabling non-default options like CA pinning — can reintroduce [vulnerabilities akin to custom implementations](https://www.blackduck.com/blog/ineffective-certificate-pinning-implementations.html).
+
+### Custom Certificate Pinning
+
+Certificate pinning can be implemented manually by overriding `TrustManager` or `HostnameVerifier` in [`HttpsURLConnection`](https://developer.android.com/reference/java/net/HttpURLConnection). This approach is highly error-prone, as it often involves directly inspecting server-sent certificates via `getPeerCertificates()`, which returns unvalidated chains that [be manipulated by attackers](https://www.blackduck.com/blog/ineffective-certificate-pinning-implementations.html).
```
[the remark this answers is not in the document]
cpholguera: This text is going to be replaced with the suggestion from below so I'll resolve

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
titze: @sushi2k What is the reason for this change proposal? Imo, "MITM" is a defined term, which I don't think we should redefine?

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -88,9 +88,17 @@ The default configuration for apps targeting Android 6.0 (API level 23) and lowe
...
+Certificate pinning is a critical security mechanism employed in Android applications to safeguard against man-in-the-middle (MITM) attacks by ensuring that the app communicates exclusively with servers possessing predefined cryptographic credentials.
+
+While effective when implemented correctly, insecure implementations potentially enable attackers to read and modify all communication. See @MASWE-0047 for more details.
+
+Various approaches exist, depending on the API level of the app, and on the used libraries. In the following, the most common ones are briefly highlighted.
+
+#### Pinning via Network Security Configuration (API 24+)
+
+The Network Security Configuration can be used to pin [declarative certificates](https://developer.android.com/training/articles/security-config.html#CertificatePinning) to specific domains. This is done by providing a `<pin-set>` in the Network Security Configuration, which is a set of digests (hashes) of the public key (`SubjectPublicKeyInfo`) of the corresponding X.509 certificate.
```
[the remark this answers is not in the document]
titze: > It applies to all network traffic within the app

this is not completely true, the NSC only applies to connections done through the Android stack. Communication via low level APIs (e.g. from native code) do not take this into account.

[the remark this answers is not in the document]
titze: I changed this to:

>The **Network Security Configuration (NSC)** is the preferred and recommended way to implement certificate pinning in Android, as it provides a declarative, maintainable, and secure approach without requiring code changes. It applies to all network traffic managed by the Android framework within the app, including `HttpsURLConnection`-based connections and `WebView` requests (unless a custom `TrustManager` is used). For communication from native code, NSC does not apply, and other mechanisms need to be considered.

cpholguera:
Excellent, thanks for the updates! I'm reviewing the tests now and will post my suggestions soon.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Steps
+
+1. Set up an intercepting proxy, for example @MASTG-TOOL-0077 or @MASTG-TOOL-0097.
+2. Install the application on a device connected to that proxy, and intercept the communication.
+3. Extract all domains which were intercepted.
+
+## Observation
+
+The output should contain a list domains, for which the interception was successful.
```
cpholguera, line 25:
```suggestion
The output should contain a list of domains for which the interception was successful.
```

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Evaluation
+
+The test case fails if any relevant domain was intercepted.
```
cpholguera, line 29: What about...?
 
```suggestion
The test fails if any **security-sensitive domain** was intercepted.
```

We could add a new prerequisite and link to it here. As in https://github.com/OWASP/owasp-mastg/blob/master/prerequisites/identify-security-relevant-contexts.md

For example (**REVIEW, AI generated**):

### Identifying Security-Sensitive Relevant Domains

#### **What is a Security-Sensitive Relevant Domain?**

A relevant domain is any endpoint where failing to enforce certificate pinning could expose sensitive data or compromise security. This typically includes:
- **Authentication endpoints** (e.g., login APIs, OAuth/token exchanges)
- **Endpoints handling sensitive user data** (e.g., financial transactions, personal data transfers)
- **Endpoints that initiate encrypted tunnels** (e.g., VPN, secure messaging services)
- **Internal API endpoints** used for secure app-server communication

#### **What is NOT a Security-Sensitive Relevant Domain?**

- **Third-party CDNs or analytics services** that the app does not control

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -119,6 +127,59 @@ If at least one of the pinned digests matches, the certificate chain will be con
...
+
+It's also possible to implement pinning in [native code](https://developer.android.com/ndk) (C/C++/Rust). By embedding or dynamically verifying certificates within compiled native libraries (`.so` files), you can increase the difficulty of bypassing or modifying the pinning checks via typical APK reverse engineering.
+
+That said, this approach requires significant security expertise and a careful design to manage certificates or public key hashes in native space. Maintenance and debugging also typically become more complex.
+
+#### Pinning in Cross-Platform Frameworks
+
+Cross-platform frameworks like Flutter, React Native, Cordova and Xamarin might require special considerations. Depending on the framework one of the following can apply:
+
+- The framework might support NSC. This is the case for Flutter apps on Android, but the NSC needs to be enabled in the `AndroidManifest`. See the [Flutter documentation](https://docs.flutter.dev/release/breaking-changes/network-policy-ios-android#migration-guide) on how to enable the network policy.
```
cpholguera, line 175: It seems this "feature" as well as the support for `android:usesCleartextTraffic="true"` in the manifest was reverted in Flutter 2.2: https://docs.flutter.dev/release/breaking-changes#reverted-change-in-2-2

According to this comment it's still kind of there: https://github.com/flutter/flutter/issues/106678#issuecomment-1169127260 but I'm not sure what's going on.

It looks like the way to do pinning in Flutter is via `securityContext.setTrustedCertificatesBytes` or `securityContext.setTrustedCertificates`

See https://api.flutter.dev/flutter/dart-io/SecurityContext/setTrustedCertificatesBytes.html

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0241.md
```diff
@@ -0,0 +1,26 @@
...
+id: MASTG-TEST-0241
+type: [static]
+weakness: MASWE-0047
+---
+
+## Overview
+
+Apps can configure expiration dates for pinned certificates in the [Network Security Configuration]("../../../Document/0x05g-Testing-Network-Communication.md#certificate-pinning"). After the expiration date, the pin is no longer used and all installed CAs for that domain are trusted.
+
+The goal of this test is to check if any expiration date is in the past.
```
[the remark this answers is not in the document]
titze: sounds good, thanks!

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0240.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Missing Certificate Pinning in Code
```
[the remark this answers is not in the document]
titze: I created a task, this PR is getting quite big.
https://github.com/OWASP/owasp-mastg/issues/3183

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Missing Certificate Pinning in Network Traffic
```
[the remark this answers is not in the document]
titze: @cpholguera not sure I understand what you mean. The link you mention explains how to _deactivate_ pinning using Frida etc. How would that help identifying if pinning is done / done correctly?

What could be (somewhat) helpful, is to try the Frida scripts to see where it attaches, to then statically analyse those locations (relating to the to-be-created test for #3183).

Am I missing something?

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
...
+
+Since statically finding all of the locations where certificate pinning is performed might not be feasible, this test case uses dynamic analysis to observe all connections the app makes.
+
+The goal of this test case is to dynamically check if the connection to a server can be intercepted using a [Man-in-the-Middle attack]("../../../Document/0x04f-Testing-Network-Communication.md#mitm-attack). If this is possible, it means that the certificate is not pinned correctly or not pinned at all.
+
+## Steps
+
+1. Set up an intercepting proxy, for example @MASTG-TOOL-0077 or @MASTG-TOOL-0097.
+2. Install the application on a device connected to that proxy, and intercept the communication.
+3. Extract all domains which were intercepted.
```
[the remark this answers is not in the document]
titze: @cpholguera which clues do you have in mind? 
The way I did this before was to just look at the intercepting proxy, logcat from the device never added anything too useful for this particular test.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
titze: @cpholguera I agree this would benefit a lot from a definition. Maybe sth to discuss in our next meeting, I don't think this is easy to define.

Imo, security-sensitive depends a lot on the app context, but also on the test / weakness itself. I could imagine certificate pinning being mandatory for L2 for specific domains, but it certainly shouldn't be mandatory for all domains for L1.

In the AI generated text, the 3rd party or analytics service is already not ideal imo. Whilst I agree with this for certificate pinning, I don't think analytics services are by default not security sensitive. E.g., thinking about GDPR, I can imagine personal identifiers being sent there, which need to be handled with care and are security sensitive...

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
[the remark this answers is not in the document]
cpholguera: Doesn't it output a `javax.net.ssl.SSLHandshakeException` / `java.security.cert.CertPathValidatorException` when pinning fails due to the presence of the MITM?
```

File: tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md (hunk 17 above)
Comment (cpholguera) [reply tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md:29]: We leave this for a separate PR.

**Review by cpholguera — CHANGES_REQUESTED** [review]

**Comment by cpholguera** [comment]
@titze this is almost ready, just a couple of suggestions and we can merge. Thank you!

File: tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md (hunk 22 above)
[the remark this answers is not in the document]
Comment (titze) [reply tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md:21]: The log message largely depends on how this is implemented. There might be some, but if the app follows best practise, it would not give an attacker any information in the logs.
The log message itself could also mean very different things. Does it mean the connection succeeds even in the precence of a Mitm, or that it was not established.
Imo, the real test is in the proxy itself: were you able to intercept, or not.

File: Document/0x05g-Testing-Network-Communication.md [hunk 23]
```diff
@@ -88,9 +88,17 @@ The default configuration for apps targeting Android 6.0 (API level 23) and lowe
...
+[Certificate pinning](0x04f-Testing-Network-Communication.md/#restricting-trust-identity-pinning) can be employed in Android apps to safeguard against Machine-in-the-Middle (MITM) attacks by ensuring that the app communicates exclusively with remote endpoints possessing specific identities.
+
+While effective when implemented correctly, insecure implementations potentially enable attackers to read and modify all communication. For more general details on pinning, refer to @MASWE-0047.
+
+Several approaches to certificate pinning exist, depending on the app's API level and the libraries used. Below, we highlight the most common methods. For a deeper dive into the specific implementations, see ["Deep Dive into Certificate Pinning on Android"](https://securevale.blog/articles/deep-dive-into-certificate-pinning-on-android/).
```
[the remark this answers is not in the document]
titze: @cpholguera aren't those considerations too broad? If no signature checks and other runtime protections are in place, I can always modify the app, and modify parts to make it less secure.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
...
+id: MASTG-TEST-0242
+type: [network]
+weakness: MASWE-0047
+---
+
+## Overview
+
+There are multiple ways an application can implement certificate pinning, including via the Android Network Security Config, custom TrustManager implementations, third-party libraries, and native code. Since some implementations might be difficult to identify through static analysis, especially when obfuscation or dynamic code loading is involved, this test uses network interception techniques to determine if certificate pinning is enforced at runtime.
+
+The goal of this test case is to observe whether a [MITM attack]("../../../Document/0x04f-Testing-Network-Communication.md#mitm-attack) can intercept HTTPS traffic from the app. A successful MITM interception indicates that the app is either not using certificate pinning or implementing it incorrectly.
```
cpholguera, line 13:
```suggestion
The goal of this test case is to observe whether a [MITM attack]("../../../Document/0x04f-Testing-Network-Communication.md#intercepting-network-traffic-through-mitm) can intercept HTTPS traffic from the app. A successful MITM interception indicates that the app is either not using certificate pinning or implementing it incorrectly.
```

Document/0x05g-Testing-Network-Communication.md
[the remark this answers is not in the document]
cpholguera: That's right, and intentional. This is to let the reader know that it can and will be bypassed (in a pentest, or by an attacker if they really want to look at the traffic; if other protections are in place, they will also try/have to remove them first).

We can clarify this further, of course, what would you suggest?

Document/0x05g-Testing-Network-Communication.md
```diff
@@ -119,6 +127,50 @@ If at least one of the pinned digests matches, the certificate chain will be con
...
+
+That said, this approach requires significant security expertise and a careful design to manage certificates or public key hashes in native space. Maintenance and debugging also typically become more complex.
+
+#### Pinning in Cross-Platform Frameworks
+
+Cross-platform frameworks like Flutter, React Native, Cordova, and Xamarin often require special considerations for certificate pinning, as they may not use the same network stack as native apps. For example, Flutter relies on its own Dart `HttpClient` (with BoringSSL) instead of the platform's networking stack, while Cordova makes network requests through JavaScript in a WebView. As a result, pinning behavior varies—some frameworks provide built-in configuration options, others rely on third-party plugins, and some offer no direct support but allow manual implementation via APIs. Understanding how a framework handles networking is crucial for ensuring proper pinning enforcement.
 ### Security Provider
```
cpholguera, line 174:
```suggestion

### Security Provider
```

Document/0x05g-Testing-Network-Communication.md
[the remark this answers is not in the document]
titze: If that's intentional, I like your proposal. I just wanted to point out that - if I am able to modify the application, I can ofc make it less secure. But it probably is a good thing to repeat that to the reader :)

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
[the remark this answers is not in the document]
cpholguera: I think you're right, but I also think it may be helpful to at least mention this as a testing tip, as the original test did. If you're trying to intercept and you get these log messages, that can be useful info. Also, if I'm not mistaken, some of these logs are coming from the system. Also, if the app doesn't follow best practices, even better for you as a tester.

I suggest we add it as a test tip, and once the demos are in place, if it's not helpful, we can remove it.

tests-beta/android/MASVS-NETWORK/MASTG-TEST-0242.md
```diff
@@ -0,0 +1,29 @@
...
+
+There are multiple ways an application can implement certificate pinning, including via the Android Network Security Config, custom TrustManager implementations, third-party libraries, and native code. Since some implementations might be difficult to identify through static analysis, especially when obfuscation or dynamic code loading is involved, this test uses network interception techniques to determine if certificate pinning is enforced at runtime.
+
+The goal of this test case is to observe whether a [MITM attack]("../../../Document/0x04f-Testing-Network-Communication.md#intercepting-network-traffic-through-mitm) can intercept HTTPS traffic from the app. A successful MITM interception indicates that the app is either not using certificate pinning or implementing it incorrectly.
+
+If the app is properly implementing certificate pinning, the MITM attack should fail because the app rejects certificates issued by an unauthorized CA, even if the CA is trusted by the system.
+
+## Steps
+
+1. Set up an intercepting proxy, for example @MASTG-TOOL-0077 or @MASTG-TOOL-0097.
```
cpholguera, line 19:
```suggestion
1. Set up an interception proxy, see @MASTG-TECH-0011.
```

tests/android/MASVS-NETWORK/MASTG-TEST-0022.md
```diff
@@ -7,6 +7,9 @@ platform: android
 title: Testing Custom Certificate Stores and Certificate Pinning
 masvs_v1_levels:
 - L2
+status: deprecated
+covered_by: [MASTG-TEST-0x22-1,MASTG-TEST-0x22-2,MASTG-TEST-0x22-3]
```
cpholguera, line 11:
```suggestion
covered_by: [MASTG-TEST-0242, MASTG-TEST-0243, MASTG-TEST-0244]
```

cpholguera (approved):
Thanks a lot @titze!!

PR #3036, opened 2024-11-05
Port MASTG-TEST-0015 to v2

### LINKED ISSUE #2948: MASTG v1->v2 MASTG-TEST-0015: Testing the Purposes of Keys (android)

- **Title:** Testing the Purposes of Keys
- **ID:** MASTG-TEST-0015
- **Link:** https://mas.owasp.org/MASTG/tests/android/MASVS-CRYPTO/MASTG-TEST-0015/
- **Platform:** android
- **MASVS v1:** ['MSTG-CRYPTO-5']
- **MASVS v2:** ['MASVS-CRYPTO-2']

nmsa:
Thank you for submitting a Pull Request to the OWASP MASTG. Please make sure that:

- [X] You verified/tested the effectiveness of your contribution (e.g.: is the code really an effective remediation? Please verify it works!)

This PR closes #2948.

@cpholguera, created a draft for us to discuss when possible
 
Most likely needs to be broken down in some atomic tests, according to what is being checked now:
1. For asymmetric keys is the key pair being used for mutiple activities (e.g. signatures vs encryption)
2. For symmetric keys, are those being used for multiple business purposes?  (hard to automate)

Case (1.) is easy to automate, (2.) is not so much.

Then, there are some other checks to be discussed:
- are all keys used according to the purpose defined during its creation? (it is relevant to KeyStore keys, which can have KeyProperties defined) <-- the APIs ensure that this is done properly, maybe can be removed
- is cryptography used according to its business purpose? <-- To be discussed if it is really different from the remaining

### DIFF
--- a/tests-beta/android/MASVS-CRYPTO/MASTG-TEST-0x15.md
+++ b/tests-beta/android/MASVS-CRYPTO/MASTG-TEST-0x15.md
@@ -1,0 +1,59 @@
+---
+platform: android
+title: Testing the Purposes of Keys
+id: MASTG-TEST-0x15
+type: [static, dynamic]
+weakness: MASWE-0012
+---
+
+## Overview
+
+In general, ["a single key or key pair shall be used for only one purpose (e.g., encryption, integrity
+authentication, key wrapping, random bit generation, or digital signatures)"](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf).
+Reusing the same key or key pair for mutiple purposes may weaken its security or increase the impact of one key being compromised.
+
+This does not apply to the cases where the same process can provide
+multiple services e.g., when a single digital signature provides integrity
+authentication and source authentication.
+Check ["NIST.SP.800-57pt1r5"](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf) for details.
+
+## Steps
+
+1. Use either @MASTG-TECH-0014 or @MASTG-TECH-0033 to identify all the where cryptography is used. 
+You can look for uses of:
+    - classes `Cipher`, `Mac`, `MessageDigest`, `Signature`
+    - interfaces `Key`, `PrivateKey`, `PublicKey`, `SecretKey`
+    - functions `getInstance`, `generateKey`
+    - exceptions `KeyStoreException`, `CertificateException`, `NoSuchAlgorithmException`
+    - classes importing `java.security.*`, `javax.crypto.*`, `android.security.*`, `android.security.keystore.*`
+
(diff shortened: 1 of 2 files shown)


cpholguera:
I like how you split them and I agree with the mapping to "MASWE-0012: Insecure or Wrong Usage of Cryptographic Key". So we can have 2 tests here at least:

1. For asymmetric keys is the key pair being used for multiple activities (e.g. signatures vs encryption)
  - key purpose (can be read from the properties of the key)
2. For symmetric keys, are those being used for multiple business purposes? (hard to automate)
  - the typical "key reuse case", right? Dangerous especially for ECB if I'm not mistaken.
  - Can we test by collecting all uses of the same key with Frida and inspecting the reversed code? e.g. The observation can contains all the uses of each key with backtraces to know the locations. In the evaluation we'd suggest to reverse engineer those places and try to determine the "business purpose". 


See https://crypto.stackexchange.com/questions/33751/aes-key-reuse-and-guessing-the-key

(Ideally we'd find another source from NIST for example)

> The reuse of the key is not a problem as long as you have chosen a mode of operation that is appropriate for your application (and of course, if you have used it properly). This is because the mode of operation will "pre-mix" your clear text data in a way that, even before encryption, two identical block of data you feed to the actual AES algorithm will never be identical (or at least, not more often than by random chance).

nmsa:
Addressed your comments according to what we discussed live yesterday.

cpholguera (changes requested):
Some general feedback after a quick review:

* Wrong title capitalization
* Profiles missing in the new tests → add L2
* Overview seems pretty generic and it's repeated. It seems that this info could go into the weakness (MASWE-0012)

  * Instead, the test overview should focus on *how the aspect of the weakness we're testing for can happen*, including specific APIs, etc.
  * Use official language from Android and NIST

    * e.g. here it refers to [[key authorizations](https://developer.android.com/privacy-and-security/keystore#KeyUseAuthorizations)](https://developer.android.com/privacy-and-security/keystore#KeyUseAuthorizations)
  * Some potentially missing APIs worth mentioning/using

    * [https://developer.android.com/reference/android/security/keystore/KeyProtection#getPurposes()](https://developer.android.com/reference/android/security/keystore/KeyProtection#getPurposes%28%29)
    * Specific `KeyProperties` (e.g. [KeyProperties.PURPOSE\_DECRYPT](https://developer.android.com/reference/android/security/keystore/KeyProperties#PURPOSE_DECRYPT))
* Steps should be aligned with similar recent tests

  * e.g. they mention many APIs now, and that should be in the overview
  * I'm not sure if the split of APIs between symmetric and asymmetric crypto is correct
* `## References` should be removed. Any references should go into the metadata refs if not already inlined in the text

cpholguera:
Done in https://github.com/OWASP/mastg/pull/3525 Thanks @serek8!

PR #3130, opened 2025-01-25
Update MASTG-TEST-0229 with info about False Positives

cpholguera:
Info about FPs was missing.


serek8 (approved):
Looks good to me. I like `Use of Memory Safe Languages` and `Compiler Optimizations` sections because they describe real world examples with Flutter and ReactNative. Without them I think that we would get a lot of issues/questions about these use cases.

Issue #2858, opened 2024-08-02
Add Additional CWE mappings to MASWE


cpholguera:
MASWE supports CWE mappings already:

https://github.com/search?q=repo%3AOWASP%2Fowasp-mastg%20%22cwe%3A%22&type=code

For example, in MASWE-0041:

```yaml
mappings:
  masvs-v1: [MSTG-AUTH-1]
  masvs-v2: [MASVS-AUTH-2]
  cwe: [603, 307, 287]
```

Review the suggestions below and add the remaining missing mappings to the rest of the MASWE.

### Discussed in https://github.com/OWASP/owasp-mastg/discussions/2857

<div type='discussions-op-text'>

<sup>Originally posted by **poffo-mobisec** August  2, 2024</sup>
I love the introduction of Weaknesses in mobile security. It was missing and it is brilliant. But let's go straight to the point.

Nowadays most of enterprises have standardized systems and works with CWE.
Have you considered relate each MASWE to a CWE, to ease the risk management and company integration?

I think this could give a lot of extra value to the project, allowing MASWE to be very specific on mobile weaknesses but at the same time bring compatibility with nowadays market.

THE FOLLOWING CONTENT IS AI-GENERATED, so this work would absolutlely need a check, but it gives the idea of the result:

| MASWE ID       | MASWE Title                                              | Relevant CWE ID | CWE Title                                           |
|----------------|----------------------------------------------------------|-----------------|----------------------------------------------------|
| MASWE-0001     | Insertion of Sensitive Data into Logs                    | CWE-532         | Insertion of Sensitive Information into Log File   |
| MASWE-0002     | Sensitive Data Stored With Insufficient Access Restrictions | CWE-200         | Exposure of Sensitive Information to an Unauthorized Actor |
| MASWE-0003     | Sensitive Data Remains in App Backups                    | CWE-212         | Improper Cross-boundary Removal of Sensitive Data  |
| MASWE-0004     | Unencrypted Sensitive Data Stored in Non-Volatile Memory | CWE-311         | Missing Encryption of Sensitive Data               |
| MASWE-0005     | Insecure Data Storage in Shared Preferences              | CWE-922         | Insecure Storage of Sensitive Information          |
| MASWE-0006     | Insecure Data Storage in SQL Databases                   | CWE-312         | Cleartext Storage of Sensitive Information         |
| MASWE-0007     | Insecure Data Storage in External Storage                | CWE-922         | Insecure Storage of Sensitive Information          |
[...]

PR #3150, opened 2025-02-13
Update CWE mapping on MASWE elements of MASVS-PRIVACY

truerick:
- Update all CWE IDs on MASWE elements of MASVS-PRIVACY

This PR is related to issue #2858 .


cpholguera:
Unfortunately we cannot consider these additions since CWE-213 is not related to privacy policies. It's about the developer's security policy. The currently referenced CWE is adequate.

https://cwe.mitre.org/data/definitions/213.html

PR #3158, opened 2025-02-14
Add Deprecation Warnings about EncryptedFile and EncryptedSharedPreferences Android classes

AndrewScull:
EncryptedSharedPreferences is from the deprecated [1] Jetpack security crypto library and, as such, should be avoided in discussion of data storage options on Android to avoid confusion.

[1] -- https://developer.android.com/privacy-and-security/cryptography#jetpack_security_crypto_library


cpholguera:
Hello @AndrewScull, the library was last updated on January 29, 2025.

https://developer.android.com/jetpack/androidx/releases/security

AndrewScull:
There are non-crypto package in the library but the crypto package only contains deprecated classes. See the package at the commit corresponding to the latest update that you linked https://android.googlesource.com/platform/frameworks/support/+/8d4dee36b185eb647753338de4e161bebaeff24d/security/security-crypto/src/main/java/androidx/security/crypto

Here is the commit that deprecated it https://android.googlesource.com/platform/frameworks/support/+/3724241aab8a48dbfd5cc147869350fd6511bb97

cpholguera:
I see, interesting. Thanks for sharing, it seems that not only is this class marked as deprecated, but so is `EncryptedFile`. 

However, Google still recommends it in the security documentation, and the API reference doesn't reflect that.

https://developer.android.com/privacy-and-security/security-best-practices#sharedpreferences

https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences


Let me ask them directly and I'll get back to you when I have new information. I'll keep the PR open for now. Note that if we remove these references/recommendations, we'll need to explain it and also offer alternatives.

cpholguera:
Google responded that this library was not actively maintained, hence it has been deprecated.  Devs can still do this directly, but Google is considering how to replace it in the future to make it easier.

Deprecation is currently indicated here: https://developer.android.com/privacy-and-security/cryptography

We're going to keep the recommendation to the library but with a disclaimer.

@TheDauntless is gonna help you with the PR.

cpholguera (approved):

Issue #3156, opened 2025-02-14
GitHub Actions Review & Optimization (build-android-demos.yml)


cpholguera:
### Discussed in https://github.com/OWASP/owasp-mastg/discussions/3155

<div type='discussions-op-text'>

<sup>Originally posted by **javier-ruiz-b** February 14, 2025</sup>
We could improve the readability and speed times of some of the GitHub actions without much increase in their complexity. 

I'd like to try improving the Android workflow and see how it goes :smile: </div>

PR #3173, opened 2025-02-22
fix: caching a demo would cause other demos to use a modified Android…

javier-ruiz-b:
Already discussed with @cpholguera.

Reduced build time to 3m 21s :partying_face: 

Corrects bug introduced with the latest closed issue #3156


cpholguera (approved):
This is amazing, thanks again!!

PR #3185, opened 2025-02-28
Use ./gradlew assembleRelease to decrease the Android demo APK size

### LINKED ISSUE #3159: Optimize / Reduce final Demo APK File size

javier-ruiz-b:
assembleRelease seems to decrease the APK size, but it also increases build time:

- 3m 21s  -> 5m 10s
- 24,9 -> 18,7MB 

@cpholguera What do you prefer, smaller APK size or lower build time?

image

This PR closes #3159


cpholguera:
I see, I think we'll keep the speed. The size really doesn't seem to improve that much in comparison. The reduction for the zipped version is still under 2MB.

But thanks for trying it out!

PR #3316, opened 2025-06-06
Fix tags

TheDauntless:
This fixes the tags shown at the top of tests, which disappeared a while ago.

### Background info

mkdocs-material implemented a rewrite of the tags functionality in 9.6.0, which is used in the docker build command. For local runs, the requirements.txt weren't necessarily respected and I was running 9.5.x myself.

In the rewrite, the on_page_markdown trigger runs on priority -50 and should run last. Our add-tags on_page_markdown function also ran at -50, so the order is (probably?) undefined. By making it -49 we can be sure it runs before the mkdocs-material/tags plugin.


cpholguera (approved):
Thanks a lot Jeroen!

PR #3350, opened 2025-06-21
Add Github statistics to github-based tools

TheDauntless:
## Adds a github image to each tool hosted on github (both light and dark mode)

image
image

I like these two themes, but it's very easy to choose another theme. Available themes: https://github.com/anuraghazra/github-readme-stats/blob/master/themes/README.md#repo-card

I've created a custom vercel app as explained on the github-readme-stats repo.

## Removes the 'available at' at the top of each tools page

Design-wise, it took up a lot of unnecessary space. Tools that are not hosted on github don't have a clickable image now, but I suggest updating those tools to use the format like https://mas.owasp.org/MASTG/tools/ios/MASTG-TOOL-0118/ (ie, starting the first paragraph with the name of the tool and making that a link).

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -35,19 +35,19 @@
 /* Default: hide both */
 img.light-img,
 img.dark-img {
-  display: none !important;
+  display: none;
 }
 
 /* Show light image only in light mode */
 [data-md-color-scheme="default"] img.light-img,
 [data-md-color-scheme="light"] img.light-img {
-  display: block !important;
+  display: block;
 }
 
 /* Show dark image only in dark mode */
 [data-md-color-scheme="slate"] img.dark-img,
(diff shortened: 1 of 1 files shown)


cpholguera:
Does this allow you to add the latest release? That'd be interesting

TheDauntless:
@cpholguera  Do you want the latest release on the card? Or as a datapoint in the mastg? The first one is doable with some custom scripting the second isn't possible this way, because it's just an image as far as the MASTG is concerned.

cpholguera:
Look, this is what I was trying to do some time ago, please take a look and let me know what you think:

https://github.com/OWASP/owasp-mastg/blob/master/src/scripts/tools_healthcheck.py

cpholguera (approved):
Very nice!!

PR #3351, opened 2025-06-21
Add lightbox and mark specific images as no-lightbox

TheDauntless:
## Add lightbox plugin

* Uses https://blueswen.github.io/mkdocs-glightbox
* Custom css class for disabling lightbox: `no-lightbox`

## Marked various images as no-lightbox

* It's on by default, but doesn't make sense for many of the 'website' images
* It will also automatically disable itself if the image is a link

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -17,7 +17,7 @@ hide:
 </div>
 
 <center>
-<img style="padding: 10px; max-width: 250px" src="assets/logo_circle.png" />
+<img style="padding: 10px; max-width: 250px" src="assets/logo_circle.png"  class="no-lightbox"/>
 </center>
 
 </div>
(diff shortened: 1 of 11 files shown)


docs/index.md
```diff
@@ -17,7 +17,7 @@ hide:
 </div>
 
 <center>
-<img style="padding: 10px; max-width: 250px" src="assets/logo_circle.png" />
+<img style="padding: 10px; max-width: 250px" src="assets/logo_circle.png"  class="no-lightbox"/>
```
cpholguera:
```suggestion
<img style="padding: 10px; max-width: 250px" src="assets/logo_circle.png" class="no-lightbox"/>
```

cpholguera:
I deployed the site for testing on mobile. I like it, but I noticed a few issues:

- When you click on an image, it opens, but you cannot zoom in.
- You can easily navigate to the next image and lose context.
- If you're on an image and use the "go back" gesture, you go back in your browsing history instead of closing the image.

TheDauntless:
* zooming: Works well on ipad, a little bit less on a smartphone, but it does work. We could potentially disable the lightbox if the screen is smaller than X pixels?
* Scrolling: Not possible to disable without a hack, but it's tricky since we're using the extension. Feels like a minor issue?
* The back-button is also not possible without hacking something together.

Are you fine with these limitations, or do you want me to search for a different lightbox implementation? I can pretty easily implement the 1st item, which would solve nr3 as well?

cpholguera:
I'll think about it, let's discuss the options next week. The less hacky, the better.

TheDauntless:
As a poc:

* No lightbox on mobile (<768 px)
* Single item per click. It's still "a hack" but one I got from the official glightbox project author
* Works on resize too, switching between lightbox and no lightbox

cpholguera:
Closing since the files have moved.

See https://github.com/OWASP/mas-website/issues/9

PR #3368, opened 2025-07-01
Fix metadata gathering and path handling in hooks

cpholguera:
Enhance the metadata gathering function to include component type and improve path handling in hooks by using filenames instead of paths for matching. This addresses issues with ID matching and ensures proper logging for missing frontmatter.

Currently, if there are any other .md files anywhere (e.g. in this case there's a demos/android/MASVS-CRYPTO/MASTG-DEMO-0058/node_modules/frida-java-bridge/README.md, the build will fail like this:

image


docs/hooks/add-cross-references.py
```diff
@@ -7,29 +7,33 @@
...
+            with open(file, 'r') as f:
+                content = f.read()
+                frontmatter = next(yaml.load_all(content, Loader=yaml.FullLoader))
 
-            if not id_key in frontmatter:
-                if "MASTG-TEST-02" in file:
-                    log.error(f"Missing frontmatter ID in {file}")
-                continue
+                if not id_key in frontmatter:
+                    if "MASTG-TEST-02" in file:
```
TheDauntless, line 23: No idea where this came from. Some left over debugging statements maybe?

TheDauntless (approved):
1 small comment

PR #3199, opened 2025-03-07
Cryptography

sydseter:
Thank you for submitting a Pull Request to the OWASP MASTG. Please make sure that:

- [ ] Your contribution is written in the 2nd person (e.g. you)
- [ ] Your contribution is written in an active present form for as much as possible.
- [ ] You have made sure that the reference section is up to date (e.g. please add sources you have used, make sure that the references to MITRE/MASVS/etc. are up to date)
- [ ] Your contribution has proper formatted markdown and/or code
- [ ] Any references to website have been formatted as [TEXT](URL “NAME”)
- [ ] You verified/tested the effectiveness of your contribution (e.g.: is the code really an effective remediation? Please verify it works!)

If your PR is related to an issue. Please end your PR test with the following line:
This PR closes #< insert number here >.

PR #3371, opened 2025-07-04
Update MASWE-0023 References for Asymmetric Cryptography With RSA PKCS#1 v1.5

cpholguera:


cpholguera:
already done in #3199

PR #3493, opened 2025-10-03
Update MASTG-TECH-0026.md

Diolor:
Fixed:
DeprecationWarning: The command 'explore' is deprecated. Use 'objection start' instead of 'objection explore'

### DIFF
--- a/techniques/android/MASTG-TECH-0026.md
+++ b/techniques/android/MASTG-TECH-0026.md
@@ -19,5 +19,5 @@ $ objection patchapk --source UnCrackable-Level1.apk
 # Install the patched APK on the Android phone
 $ adb install UnCrackable-Level1.objection.apk
 # After running the mobile phone, objection will detect the running frida-server through the APK
-$ objection explore
+$ objection start
 ```


sushi2k:
Thanks for adding. People who just install the "latest" objection version 1.11.0 via pip (and not install manually via main branch) will only be able to use `objection explore`, there is no `start` command.  This is also what we suggest in the tools page: https://mas.owasp.org/MASTG/tools/generic/MASTG-TOOL-0038/ on how to install objection.

I would suggest we wait to merge this PR till a new release is being published and the `start` command is also available in the official release to avoid confusion. It seems a new release is in the making: https://github.com/sensepost/objection/issues/756

Diolor:
Uh good catch @sushi2k - marked it as blocked

TheDauntless:
@Diolor  How about we update the objection techniques to include a remark about the current state of objection, and that the techniques and commands are meant for the latest version from the github until v2 is released. When v2 is released, we will remove the notification and update all the commands (including the internal objection commands, which have also been changed a bit). That way we can also close this PR :)

I can do it, or you can do it yourself in this PR if you want.

Diolor:
Let's do that @TheDauntless. :) Could you please take a look at the 4 pages I have touched with my PR and validate if anything else is worth mentioning in those notes aside from the explore/start command change? You might have a more solid understanding of current objection's state than I do.

I will change this PR to the notes style in the meantime


Update:

I drafted the info box

techniques/android/MASTG-TECH-0026.md
```diff
@@ -3,6 +3,12 @@ title: Dynamic Analysis on Non-Rooted Devices
 platform: android
 ---
 
+??? info "Info about objection"
+
+    The commands below are for objection version 1.11.0. At the moment of writing this document, a version 2 of objection is expected, which would support Frida 17 and above.
```
TheDauntless:
```suggestion
    The commands below are for objection version 1.11.0, which relies on Frida < 17. To use objection, install `frida-tools==13.7.1` and use a `frida-server` lower than 17 on your device. If you want to use objection with Frida 17, you can pull the latest version from the objection repository. Note that some commands have been modified in the next release, so you may have to modify the steps below. Once the updated version has officially been released, the steps below will be updated.
```
I was thinking of something like this. Please review and adapt the others.

cpholguera (approved):

PR #3558, opened 2025-12-01
MASTG v1->v2 MASTG-TEST-0037: Testing WebViews Cleanup (android)

### LINKED ISSUE #2975: MASTG v1->v2 MASTG-TEST-0037: Testing WebViews Cleanup (android)

- **Title:** Testing WebViews Cleanup
- **ID:** MASTG-TEST-0037
- **Link:** https://mas.owasp.org/MASTG/tests/android/MASVS-PLATFORM/MASTG-TEST-0037/
- **Platform:** android
- **MASVS v1:** ['MSTG-PLATFORM-10']
- **MASVS v2:** ['MASVS-PLATFORM-2']

Diolor:
This PR closes #2975

## Description

Migrate test-0037 to v2. 
Includes a new demo which has a frida hook to inspect webview cleanup.

TODO before merge: 

- [x] Allocate IDs to TEST and DEMO

### DIFF
--- a/techniques/android/MASTG-TECH-0143.md
+++ b/techniques/android/MASTG-TECH-0143.md
@@ -3,7 +3,7 @@ title: Monitor File System Operations in WebViews
 platform: android
 ---
 
-You can monitor any file system operations in the WebView storage directory using various techniques: 
+You can monitor any file system operations in the WebView storage directory using various techniques:
 
 Use @MASTG-TECH-0033 to monitor file system operations in the `/data/data/<app_package>/app_webview/` directory. Regardless of whether the app uses these APIs directly, WebViews may use them internally when rendering content (e.g., JavaScript code using `localStorage`). So tracing calls to APIs such as `open`, `openat`, `opendir`, `unlinkat`, etc., can help identify file operations in the WebView storage directory.
 


demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/MASTG-DEMO-00de.md
```diff
@@ -0,0 +1,36 @@
...
+3. Run `run.sh` to spawn the app with Frida
+4. Click the **Start** button in the app
+5. Wait for the Frida script to capture the WebView cleanup calls
+6. Stop the script by quitting the Frida CLI
+
+{{ hooks.js # run.sh }}
+
+### Observation
+
+The output shows all instances of `deleteAllData()` of `WebStorage` called at runtime. A backtrace is also provided to help identify the location in the code.
```
cpholguera: The output doesn't show any calls to `deleteAllData`. Only `getInstance`.

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/MASTG-DEMO-00de.md
```diff
@@ -0,0 +1,36 @@
...
+
+{{ output.json }}
+
+### Evaluation
+
+The test **succeeds** as the application properly cleans up all storage data from the WebView cache using the `WebStorage` API.
+
+{{ evaluate.sh }}
```
cpholguera: Missing file?

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,55 @@
...
+title: Testing WebViews Cleanup
+id: MASTG-TEST-02te // TODO replace with real ID
+type: [dynamic, manual]
+weakness: MASWE-0118
+profiles: [L1, L2]
+prerequisites:
+- identify-sensitive-data
+---
+
+## Overview
```
cpholguera, line 14: Any chance we could put some of this content in an existing or new KNOW?

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,55 @@
+---
+platform: android
+title: Testing WebViews Cleanup
+id: MASTG-TEST-02te // TODO replace with real ID
+type: [dynamic, manual]
+weakness: MASWE-0118
+profiles: [L1, L2]
+prerequisites:
+- identify-sensitive-data
+---
```
cpholguera, line 12: Could you please link to an existing or add a new BEST?

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/MASTG-DEMO-00de.md
Diolor: typo. tnx!

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/MASTG-DEMO-00de.md
Diolor: true! I'll debug 🤔

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
Diolor: Is it preferred to **move** or **copy** those? `MASTG-KNOW-0018/#webviews-cleanup` will do.

I could possibly move some content away from here to BEST **and** KNOW. Those might have some overlapping content. Will try to see what I can do.

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
Diolor: Like...with the "opposite" content? E.g., title `Cleanup Webviews`?

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
Diolor: Moved content to KNOW and BEST.
- KNOW-0018 has information about the storage areas. It is referenced from this overview.
- BEST-00be has recommendations about cache and clean up. It also references KNOW-0018.
- This test references BEST-00be

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,40 @@ Data from other apps accessible via content providers (if the app has any and th
...
+- other files stored locally backed by the Origin Private File System (OPFS), such as the SQLite Wasm database
+
+> Note: WebSQL was deprecated in Android when version 15 was released. To learn more about the World Wide Web Consortium (W3C) recommendations, visit the [deprecation note](https://developer.android.com/about/versions/15/deprecations#websql-webview)
+
+### Clearing methods
+
+Clearing methods can be generic or granular and vary depending on the storage area that should be purged or the application's functionality.
+
+- **Cached files**: [`WebView.clearCache(includeDiskFiles = true)`](https://developer.android.com/reference/android/webkit/WebView#clearCache(boolean)) method can be called to delete both the RAM cache and files stored locally (i.e., images, JS, CSS). This is a per-application operation that clears the cache for all WebViews.
+- **WebStorage APIs**: [`WebStorage.clearAllData()`] Clears DOM storage (local and session storage), Web SQL Database, and HTML5 Web Storage APIs, including IndexedDB.
```
cpholguera: it's `deleteAllData`, right? 

```suggestion
- **WebStorage APIs**: [`WebStorage.deleteAllData()`](https://developer.android.com/reference/android/webkit/WebStorage#deleteAllData()) clears DOM storage (local and session storage), Web SQL Database, and HTML5 Web Storage APIs, including IndexedDB.
```

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,40 @@ Data from other apps accessible via content providers (if the app has any and th
...
+- **Cookies**: [`CookieManager.removeAllCookies(ValueCallback<Boolean> ...)`] Clears all cookies.
+- **OPFS**: [`java.io.File.deleteRecursively`] Deletes the file.
+- **SQLite Wasm**: [`SQLiteDatabase.delete()`] to delete rows or [`SQLiteDatabase.deleteDatabase()`] to delete the database.
```
cpholguera: Missing links. Check capitalization.

cpholguera: OPFS?

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,40 @@ Data from other apps accessible via content providers (if the app has any and th
...
+Please note that **when you use `addJavascriptInterface`, you're explicitly granting access to the registered JavaScript Interface object to all pages loaded within that WebView**. This implies that, if the user navigates outside your app or domain, all other external pages will also have access to those JavaScript Interface objects, which might present a potential security risk if any sensitive data is being exposed through those interfaces.
 
 > Warning: Take extreme care with apps targeting Android versions below Android 4.2 (API level 17) as they are [vulnerable to a flaw](https://labs.withsecure.com/publications/webview-addjavascriptinterface-remote-code-execution "WebView addJavascriptInterface Remote Code Execution") in the implementation of `addJavascriptInterface`: an attack that is abusing reflection, which leads to remote code execution when malicious JavaScript is injected into a WebView. This was due to all Java Object methods being accessible by default (instead of only those annotated).
 
 ## WebViews Cleanup
```
cpholguera: I don't see anything wrong about the content in the test. If anything it can be updated but I'd keep it including the Firefox example.

````md
- **Initialization**: an app might be initializing the WebView in a way to avoid storing certain information by using `setDomStorageEnabled`, `setAppCacheEnabled` or `setDatabaseEnabled` from [`android.webkit.WebSettings`](https://developer.android.com/reference/android/webkit/WebSettings "WebSettings"). The DOM Storage (for using the HTML5 local storage), Application Caches and Database Storage APIs are disabled by default, but apps might set these settings explicitly to "true".

- **Cache**: Android's WebView class offers the [`clearCache`](https://developer.android.com/reference/android/webkit/WebView#clearCache(boolean) "clearCache in WebViews") method which can be used to clear the cache for all WebViews used by the app. It receives a boolean input parameter (`includeDiskFiles`) which will wipe all stored resource including the RAM cache. However if it's set to false, it will only clear the RAM cache. Check the app for usage of the `clearCache` method and verify its input parameter. Additionally, you may also check if the app is overriding `onRenderProcessUnresponsive` for the case when the WebView might become unresponsive, as the `clearCache` method might also be called from there.

- **WebStorage APIs**: [`WebStorage.deleteAllData`](https://developer.android.com/reference/android/webkit/WebStorage#deleteAllData) can be also used to clear all storage currently being used by the JavaScript storage APIs, including the Web SQL Database and the HTML5 Web Storage APIs.
  > Some apps will _need_ to enable the DOM storage in order to display some HTML5 sites that use local storage. This should be carefully investigated as this might contain sensitive data.

- **Cookies**: any existing cookies can be deleted by using [CookieManager.removeAllCookies](https://developer.android.com/reference/android/webkit/CookieManager#removeAllCookies(android.webkit.ValueCallback%3Cjava.lang.Boolean%3E)).

- **File APIs**: proper data deletion in certain directories might not be that straightforward, some apps use a pragmatic solution which is to _manually_ delete selected directories known to hold user data. This can be done using the `java.io.File` API such as [`java.io.File.deleteRecursively`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/delete-recursively.html).

**Example:**

This example in Kotlin from the [open source Firefox Focus](https://github.com/mozilla-mobile/focus-android/blob/v8.17.1/app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt#L220 "Firefox Focus for Android") app shows different cleanup steps:

```Java
override fun cleanup() {
    clearFormData() // Removes the autocomplete popup from the currently focused form field, if present. Note this only affects the display of the autocomplete popup, it does not remove any saved form data from this WebView's store. To do that, use WebViewDatabase#clearFormData.
    clearHistory()
    clearMatches()
    clearSslPreferences()
    clearCache(true)

    CookieManager.getInstance().removeAllCookies(null)

    WebStorage.getInstance().deleteAllData() // Clears all storage currently being used by the JavaScript storage APIs. This includes the Application Cache, Web SQL Database and the HTML5 Web Storage APIs.

    val webViewDatabase = WebViewDatabase.getInstance(context)
    // It isn't entirely clear how this differs from WebView.clearFormData()
    @Suppress("DEPRECATION")
    webViewDatabase.clearFormData() // Clears any saved data for web forms.
    webViewDatabase.clearHttpAuthUsernamePassword()

    deleteContentFromKnownLocations(context) // calls FileUtils.deleteWebViewDirectory(context) which deletes all content in "app_webview".
}
```

The function finishes with some extra _manual_ file deletion in `deleteContentFromKnownLocations` which calls functions from [`FileUtils`](https://github.com/mozilla-mobile/focus-android/blob/v8.17.1/app/src/main/java/org/mozilla/focus/utils/FileUtils.kt). These functions use the [`java.io.File.deleteRecursively`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/delete-recursively.html) method to recursively delete files from the specified directories.

```Java
private fun deleteContent(directory: File, doNotEraseWhitelist: Set<String> = emptySet()): Boolean {
    val filesToDelete = directory.listFiles()?.filter { !doNotEraseWhitelist.contains(it.name) } ?: return false
    return filesToDelete.all { it.deleteRecursively() }
}
```
````

File: tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md [hunk 8]
```diff
@@ -0,0 +1,32 @@
...
+best-practices: [MASTG-BEST-00be]
+prerequisites:
+- identify-sensitive-data
+---
+
+## Overview
+
+This test verifies that the application cleans up sensitive information used by WebViews.
+
+@MASTG-KNOW-0018 describes the different storage areas used by WebViews.
```
cpholguera: Let's indicate the ones we're targeting in this test as we do in other tests.

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
```diff
@@ -0,0 +1,32 @@
...
+## Steps
+
+1. Install and run the app.
+2. Navigate to the webview of the mobile app you want to ensure is appropriately cleaned up.
+3. Optionally, manually debug the webview by @MASTG-TECH-0141 to inspect the storage areas for their content.
+4. Execute a method trace (@MASTG-TECH-0033) (using e.g. @MASTG-TOOL-0001) by attaching to the running app, targeting webview cleanup APIs, and save the output.
+
+## Observation
+
+The output should list the locations in the app where webview cleanup APIs are used during the current execution.
```
cpholguera: I think "the lack of them" is very interesting in this case. The test should account for this. Are we seeing `setDomStorageEnabled`, `setAppCacheEnabled` or `setDatabaseEnabled` but not their "clearing" methods? Is the app using `localStorage.setItem` and not clearing?

Also, if an app doesn't even use WebViews this test shouldn't fail. It would not apply. If the WebView doesn't handle sensitive data the test shouldn't fail either.

```suggestion
The output should list the locations in the app where WebView cleanup APIs are used during the current execution or the lack of them.
```

demos/android/MASVS-PLATFORM/MASTG-DEMO-0082/MainActivityWebView.kt
cpholguera, line 1: This isn't needed, but we need the AndroidManifest with:

```xml
            android:name=".MainActivityWebView"
```

best-practices/MASTG-BEST-00be.md
```diff
@@ -0,0 +1,14 @@
...
+alias: android-webviews-cleanup
+id: MASTG-BEST-00be // TODO allocate real ID
+platform: android
+---
+
+Android WebViews cache data when the server responds with specific `Cache-Control` headers that instruct the browser to cache the content.
+If a WebView processes sensitive data, you should ensure that no residual information remains on the device (disk and/or RAM) once the WebView is no longer required.
+
+Prefer server-side cache prevention using headers such as `Cache-Control: no-cache` to avoid storing sensitive content.
+If server-side control is not possible, or as an supplementary control, explicitly clear the WebView cache and related data after use to reduce the risk of exposing sensitive information.
```
cpholguera: Ideally we should define the way to do it here. For example, "call `webStorage.deleteAllData()` from `onDestroy`". However, in this case the system does not guarantee `onDestroy` will always run, for example:

1. `adb shell am force-stop` tells the system to kill your app process immediately and clear its state. No lifecycle callbacks are invoked, so `onDestroy` will never be called here.
2. Swiping the app away from recents should finish the Activity and call `onDestroy`, but in practice OEMs and Android versions can treat this more like a process kill. There is no guarantee that `onDestroy` will always run before the process is killed.

Because of that, you cannot rely on `onDestroy` for security critical cleanup like erasing `WebStorage`.

You _could_:

- Clear it on the next start in addition to on destruction.
- Clear it proactively on sensitive actions, for example when the user logs out or navigates away from authenticated content.

But I don't think there's any reliable way of getting it called on destroy, so we must make this clear.

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/script.js
cpholguera, line 1: rename to script.js, hooks.js is reserved for "hooks with the frida util".

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/MASTG-DEMO-00de.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Observation
+
+The output shows all instances of `deleteAllData()` of `WebStorage` called at runtime. A backtrace is also provided to help identify the location in the code.
+
+{{ output.json }}
+
+### Evaluation
+
+The test **succeeds** as the application properly cleans up all storage data from the WebView cache using the `WebStorage` API.
```
cpholguera: Prefer passes, we typically use fails/passes.

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Evaluation
+
+The test case fails if you can find sensitive data being logged using those APIs.
```
cpholguera: Consider the previous comment. Also we should explain here or in the overview the complexity / unreliability of this. It is unreliable (both as a fix and for testing dynamically) because you cannot be sure it will be triggered.

In this case maybe the static test would be more reliable but the fix/solution isn't. Either we:

- don't include this test at all, but explain why somewhere.
- include it but mark it as "unreliable".

We have another one "similar" https://mas.owasp.org/MASTG/tests/ios/MASVS-STORAGE/MASTG-TEST-0215/ (it relies on "`isExcludedFromBackup` which exists only to provide guidance to the system about which files and directories it can exclude; it's not a mechanism to guarantee those items never appear in a backup or on a restored device.")

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/script.js
```diff
@@ -0,0 +1,17 @@
...
+  let clazz = method.classes[0].name;
+
+  Java.classFactory.loader = loader;
+
+  const WebStorageAdapter = Java.use(clazz);
+  WebStorageAdapter.deleteAllData.implementation = function() {
+    console.log('WebStorage.deleteAllData() called');
+    return this.deleteAllData();
+  }
+});
```
cpholguera: This is working better for me.

```suggestion
Java.perform(function () {
  function waitForMethods() {
    try {
      let methods = Java.enumerateMethods('com.android.webview.chromium.*!deleteAllData');
      if (methods && methods.length > 0) {
        console.log('Chromium methods found');
        console.log(JSON.stringify(methods, null, 2));

        let method = methods[0];
        let loader = method.loader;
        let clazz = method.classes[0].name;

        Java.classFactory.loader = loader;

        const WebStorageAdapter = Java.use(clazz);
        WebStorageAdapter.deleteAllData.implementation = function () {
          console.log('WebStorage.deleteAllData() called');
          return this.deleteAllData();
        };
        return;
      }
    } catch (e) {
    }

    setTimeout(waitForMethods, 500);
  }

  console.log('Waiting for chromium to be loaded');
  waitForMethods();
});

```

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/run.sh
cpholguera, line 1: You can add automation with:

flow.yaml:

```yaml
appId: org.owasp.mastestapp
---
- takeScreenshot: before
- tapOn: "Start"
- takeScreenshot: after
```

run.sh:

```sh
frida -U -f org.owasp.mastestapp -l script.js -o output.json
```

auto.sh

```sh
#!/usr/bin/env bash
set -euo pipefail

FLOW="flow.yaml"

# Start Frida and redirect stdout and stderr to file
./run.sh 2>&1 &

FRIDA_PID=$!

# Run Maestro (https://docs.maestro.dev/getting-started/installing-maestro)
maestro test "$FLOW" > auto.log 2>&1
MAESTRO_EXIT=$?

# Stop Frida when Maestro completes
kill "$FRIDA_PID" 2>/dev/null || true

exit "$MAESTRO_EXIT"
```

demos/android/MASVS-PLATFORM/MASTG-DEMO-0082/MainActivityWebView.kt
Diolor: We do because of the lifecycle (onDestroy). The alternative is to register the destroy callback within application with this boilerplate code.

This comes with the complication that `WebStorage.getInstance()` is loaded in memory only onDestroy. 
An idea I had was to use Frida to load `WebStorage.getInstance()` in memory so we can register the hook.

Even if the hook works, the snippet bellow is an overkill over the benefit of keeping the activity untouched.

```
val application: Application? = (context as? Activity)?.application

        application?.registerActivityLifecycleCallbacks(object :
            Application.ActivityLifecycleCallbacks {
            fun onActivityDestroyed() {
                WebStorage.getInstance().deleteAllData()
                application.unregisterActivityLifecycleCallbacks(this)
            }

            override fun onActivityCreated(
                activity: Activity,
                savedInstanceState: Bundle?
            ) {}

            override fun onActivityStarted(activity: Activity) {}

            override fun onActivityResumed(activity: Activity) {}

            override fun onActivityPaused(activity: Activity) {}

            override fun onActivityStopped(activity: Activity) {}

            override fun onActivitySaveInstanceState(
                activity: Activity,
                outState: Bundle
            ) {}

            override fun onActivityDestroyed(activity: Activity) {}
        })
```

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/script.js
Diolor: hm. The 500 might not work 100% for others making unstable. Let me try preload the WebStorage in apps memory via Frida.

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/MASTG-DEMO-00de.md
Diolor: good catch

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
Diolor: Will add links 👍
OPFS -> line 217

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
Diolor: I made split the structure in 2: Identify areas / clear areas. 
In `Sensitive information storage areas` I rewrote the content to be more cohesive and to the point.
In `Clearing methods` I enumerate the ways the areas can be cleaned. I will add the example here.

Did you spot any content from v1 test migrated here that it's missing?

best-practices/MASTG-BEST-00be.md
Diolor: True. I will provide a recommendation and an reliability indication.

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
Diolor: Along with the `Observation` comment, I agree, but it can be trrricky to avoid context duplication with the KNOW.

I clustered combinations of "enabled storage areas" and "cleanup". I think the test needs 4 subassements (one per storage area) to answer the initial question: "Do we clean sensitive data"?

Having 4 different tests might give an incomplete picture. Having the test as is is also incomplete.
Let's do something like subassements.

TODO: I did not find much info on relatively new SQLite Wasm.

demos/android/MASVS-PLATFORM/MASTG-DEMO-00de/script.js
Diolor: Please retry with updated `script` and `MainActivity`. If that's unreliable, paste your error here to check further...

demos/android/MASVS-PLATFORM/MASTG-DEMO-0082/MainActivityWebView.kt
cpholguera: Right, I forgot to remove the comment after I tested it the other day, sorry.

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
cpholguera: This is looking better now, thanks!

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,82 @@ Data from other apps accessible via content providers (if the app has any and th
...
+- **WebStorage APIs**: [`WebStorage.deleteAllData()`](https://developer.android.com/reference/android/webkit/WebStorage#deleteAllData()) Clears DOM storage (local and session storage), Web SQL Database, and HTML5 Web Storage APIs, including IndexedDB.
+- **Cookies**: [`CookieManager.removeAllCookies(ValueCallback<Boolean> ...)`](https://developer.android.com/reference/android/webkit/CookieManager#removeAllCookies(android.webkit.ValueCallback%3Cjava.lang.Boolean%3E)) Clears all cookies.
+- **OPFS**: [`java.io.File.deleteRecursively`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/delete-recursively.html) Deletes the file.
+- **SQLite Wasm**: [`SQLiteDatabase.delete()`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#delete(java.lang.String,%20java.lang.String,%20java.lang.String[])) to delete rows or [`SQLiteDatabase.deleteDatabase()`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#deleteDatabase(java.io.File)) to delete the database.
```
cpholguera: Needs some consistency.
```suggestion
- **Cached files**: [`WebView.clearCache(includeDiskFiles = true)`](https://developer.android.com/reference/android/webkit/WebView#clearCache(boolean)) can be called to delete both the RAM cache and files stored locally (i.e., images, JS, CSS). This is a per-application operation that clears the cache for all WebViews.
- **WebStorage APIs**: [`WebStorage.deleteAllData()`](https://developer.android.com/reference/android/webkit/WebStorage#deleteAllData()) clears DOM storage (local and session storage), Web SQL Database, and HTML5 Web Storage APIs, including IndexedDB.
- **Cookies**: [`CookieManager.removeAllCookies(ValueCallback<Boolean> ...)`](https://developer.android.com/reference/android/webkit/CookieManager#removeAllCookies(android.webkit.ValueCallback%3Cjava.lang.Boolean%3E)) clears all cookies.
- **OPFS**: [`java.io.File.deleteRecursively`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/delete-recursively.html) deletes files recursively.
- **SQLite Wasm**:
    - [`SQLiteDatabase.delete()`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#delete(java.lang.String,%20java.lang.String,%20java.lang.String[])) deletes rows
    - [`SQLiteDatabase.deleteDatabase()`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#deleteDatabase(java.io.File)) deletes the whole database.
```

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,82 @@ Data from other apps accessible via content providers (if the app has any and th
...
+```Java
+private fun deleteContent(directory: File, doNotEraseWhitelist: Set<String> = emptySet()): Boolean {
+    val filesToDelete = directory.listFiles()?.filter { !doNotEraseWhitelist.contains(it.name) } ?: return false
+    return filesToDelete.all { it.deleteRecursively() }
+}
+```
+
+## Unresponsive WebView Detection
+
+Starting on Android 10 (API level 29), apps can detect if a WebView has become [unresponsive](https://developer.android.com/about/versions/10/features?hl=en#webview-hung "WebView hung renderer detection"). If this happens, the OS will automatically call the `onRenderProcessUnresponsive` method.
```
cpholguera: is this interesting in terms of security?

Diolor: I didn't want to expand the PR scope and evaluate the content. Only moved it to make sense given the new lines. Since you ask, most likely not.

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
```diff
@@ -0,0 +1,35 @@
...
+- When [`setDomStorageEnabled`](https://developer.android.com/reference/android/webkit/WebSettings#setDomStorageEnabled(boolean)) is enabled, [`WebStorage.deleteAllData()`](https://developer.android.com/reference/android/webkit/WebView#clearFormData()) should be called.
+- When [`WebSettings.setDatabaseEnabled()`](https://developer.android.com/reference/android/webkit/WebSettings#setDatabaseEnabled(boolean)) is enabled, [`WebStorage.deleteAllData()`](https://developer.android.com/reference/android/webkit/WebView#clearFormData()) should be called.
+- When [`CookieManager.setAcceptCookie()`](https://developer.android.com/reference/android/webkit/CookieManager#setAcceptCookie(boolean)) is not explicitly set to `false` (default is set to `true`), [`CookieManager.removeAllCookies(ValueCallback<Boolean> ...)`](https://developer.android.com/reference/android/webkit/CookieManager#removeAllCookies(android.webkit.ValueCallback%3Cjava.lang.Boolean%3E)) should be called.
```
cpholguera: See https://github.com/OWASP/mastg/pull/3585/files#diff-9cc012be5cfd1157a5752e5a1983c2906aa8f52d33918aaec78362348d8593aeR163

This wasn't explicitly written in the docs yet but we are following that.

Diolor: that's good advice. I slightly rephrased to point the issue

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
serek8: I think `isExcludedFromBackup` is completely different because it provides only a guidance to the system. The API in this test actually always remove the data, right?

This test makes sense both statically and dynamically. I can see it useful in a case where a user signs out. Then, the tester could be sure that all data is deleted. 

Static test could just check if these APIs are used at all to prove that the developer is aware of the issue. To sum up, I would have two tests,

best-practices/MASTG-BEST-00be.md
serek8: It's good to point out how to secure it at the backend site but I would still put on top a way to secure it on mobile. 

Anyway, isn't `WebSettings.LOAD_NO_CACHE` enough not to cache it despite server's headers? Shouldn't it be the main _best practise_?

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
Diolor: These are some valid perspectives that your bring @serek8. Some "loud thinking" to open pandoras box ahead:

If we wish to go one step up on "being critical" we can add all possible factors of complexity/observations here: 
1. Lifecycle of data 
2. Number of webview instances. 
3. Number of storage "areas" (Android apis) 
4. State of Backend calls (cache or no cache)

In a real application we also require more confinement, such as answers to:
- Do we have one or many webviews?
- Was the "clean call" logged from DAST **relevant** to the webview, or was it randomly triggered from a different area in the app area?
- What is the lifecycle of sensitive data? User authenticated session are common, but maybe higher levels of minimization could be achieved.
- Is a static test here **competent** (=standalone)? Is a dynamic test here competent?
- Could we even in a white-box SAST, in all apps and cases prove that we can evaluate the lifecycle only with SAST without running the app? Most likely no.

So I think D-test is the **minimum** required testing which can provide competency. S-test can be very helpful but not standalone due to lack **lifecycle** proof.

Wdyt? Where do we draw the lines of confinement?

Proposal: 
1. We add the above info a best as possible in the test. 
2. We propose on minimum a D-test, we propose that a S-test can be very helpful for information gathering.

james-franco

best-practices/MASTG-BEST-00be.md
Diolor: @serek8 good question. "It depends". It might be an excessive measure though: this will disable Cache completely e.g. for  .IMG files while the sensitive data are stored in e.g. a SQL DB. Security shouldn't step on functionality when possible.

Update: I will rewrite this as best as possible and you can provide me feedback

Diolor:
@cpholguera & @serek8, please re-review. 
I tried to capture all edge cases and all points as best as possible thus I marked all conversations as resolved. Please reopen them as required or create new threads.

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,89 @@ Data from other apps accessible via content providers (if the app has any and th
...
 
-Please note that **when you use `addJavascriptInterface`, you're explicitly granting access to the registered JavaScript Interface object to all pages loaded within that WebView**. This implies that, if the user navigates outside your app or domain, all other external pages will also have access to those JavaScript Interface objects which might present a potential security risk if any sensitive data is being exposed though those interfaces.
+Please note that **when you use `addJavascriptInterface`, you're explicitly granting access to the registered JavaScript Interface object to all pages loaded within that WebView**. This implies that, if the user navigates outside your app or domain, all other external pages will also have access to those JavaScript Interface objects, which might present a potential security risk if any sensitive data is being exposed through those interfaces.
```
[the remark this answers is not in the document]
Diolor: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/OWASP/mastg/pull/3558#discussion_r2620835690)

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
```diff
@@ -0,0 +1,48 @@
...
+
+In all the above cases, if specific storage areas are enabled but cleanup methods are not called, sensitive data may be retained beyond its intended lifetime.
+
+!!! Warning
+    This test scopes out Storage areas such as [**SQLite Wasm**](https://developer.chrome.com/blog/sqlite-wasm-in-the-browser-backed-by-the-origin-private-file-system) and other APIs used by Origin Private File System (OPFS).
+
+## Steps
+
+1. Reverse engineer the app and use @MASTG-TECH-0014 to inspect WebView storage enablement APIs for the target WebView. Alternatively, install and run the app with @MASTG-TECH-0033 to target WebView storage enablement APIs and navigate to the WebView you want to evaluate.
+2. Optionally, use @MASTG-TECH-0142 to manually inspect the storage areas used by the WebView and their contents. Manually enumerate if sensitive data is present. Note that the webview could store sensitive data **after** the observation moment.
```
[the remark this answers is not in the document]
Diolor: false positive

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-02te.md
```diff
@@ -0,0 +1,48 @@
...
+3. Use @MASTG-TECH-0033 to target WebView cleanup APIs and save the output.
+
+## Observation
+
+The output should list the locations in the app where:
+
+1. The WebView enables particular storage areas.
+2. Cleanup APIs are used during the current execution, or the lack of them.
```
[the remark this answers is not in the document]
Diolor: false positive

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
+This test uses dynamic analysis to monitor the relevant API calls and file system operations. Regardless of whether the app uses these APIs directly, WebViews may use them internally when rendering content (e.g., JavaScript code using `localStorage`). So tracing calls to APIs such as `open`, `openat`, `opendir`, `unlinkat`, etc., can help identify file operations in the WebView storage directory.
+
+## Steps
+
+1. Install the app on a device (@MASTG-TECH-0056).
+2. Use @MASTG-TECH-0033 to target WebView APIs for storage and cleanup.
+3. Open the app.
+4. Use the app extensively to ensure that all relevant WebViews are covered and that sensitive data is loaded into them. Ensure you keep a list of the sensitive data you expect to be cleaned up.
+5. Close the app.
+6. Use @MASTG-TECH-0002 to pull the contents of the `/data/data/<app_package>/app_webview/` directory or simply search for the sensitive data used in the WebView within that directory.
```
Diolor: Is this for all apis and devices? This directory should also be added in KNOW.

KNOW -> BEST -> TEST
Know how something works -> Learn how to best secure it -> Assert based on previous knowledge

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
```
Diolor, line 37: This directory should also be explained in KNOW.

KNOW -> BEST -> TEST
Know how something works -> Learn how to best secure it -> Assert based on previous knowledge

Can enumerating that dir replace manual information gathering with MASTG-TECH-0142?

cpholguera: See https://github.com/OWASP/mastg/pull/3558/commits/277a655a62fa69bcb94b2ad10978150f4567ab8f

cpholguera: And https://github.com/OWASP/mastg/issues/3602

knowledge/android/MASVS-PLATFORM/MASTG-KNOW-0018.md
```diff
@@ -198,16 +198,94 @@ Data from other apps accessible via content providers (if the app has any and th
...
+
+### Clearing Stored Data
+
+Android does not provide a dedicated API to delete the Chromium profile under `app_webview`. Apps must not attempt to delete this directory directly. The only supported way to remove it is to clear the app's data, either through system settings or by calling `ActivityManager.clearApplicationUserData()`. However, this might not be desirable if the app wants to retain other user data.
+
+A more adequate approach is to clear individual storage subsystems used by WebView. These include:
+
+- **Cached Resources**: [`WebView.clearCache`](https://developer.android.com/reference/android/webkit/WebView#clearCache(boolean))(true) clears the memory and disk HTTP cache. It does not remove cookies, DOM storage, IndexedDB, OPFS, or other persistent data.
+- **WebStorage APIs**: [`WebStorage.deleteAllData`](https://developer.android.com/reference/android/webkit/WebStorage#deleteAllData()) clears LocalStorage and legacy WebSQL. It does not clear IndexedDB or OPFS.
```
Diolor: What about Session storage? Let's use the combined DOM definitiion

```suggestion
- **WebStorage APIs**: [`WebStorage.deleteAllData`](https://developer.android.com/reference/android/webkit/WebStorage#deleteAllData()) clears DOM Storage and legacy WebSQL. It does not clear IndexedDB or OPFS.
```

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
+    - [`WebSettings.setDatabaseEnabled()`](https://developer.android.com/reference/android/webkit/WebSettings#setDatabaseEnabled(boolean)) is enabled.
+- Not calling [`CookieManager.removeAllCookies(ValueCallback<Boolean> ...)`](https://developer.android.com/reference/android/webkit/CookieManager#removeAllCookies(android.webkit.ValueCallback%3Cjava.lang.Boolean%3E)) when:
+    - [`CookieManager.setAcceptCookie()`](https://developer.android.com/reference/android/webkit/CookieManager#setAcceptCookie(boolean)) is not explicitly set to `false` (default is set to `true`).
+
+This test uses dynamic analysis to monitor the relevant API calls and file system operations. Regardless of whether the app uses these APIs directly, WebViews may use them internally when rendering content (e.g., JavaScript code using `localStorage`). So tracing calls to APIs such as `open`, `openat`, `opendir`, `unlinkat`, etc., can help identify file operations in the WebView storage directory.
+
+## Steps
+
+1. Install the app on a device (@MASTG-TECH-0056).
+2. Use @MASTG-TECH-0033 to target WebView APIs for storage and cleanup.
```
Diolor:
```suggestion
2. Use @MASTG-TECH-0033 to target WebView APIs for storage enablement and cleanup.
```

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
+3. Open the app.
+4. Use the app extensively to ensure that all relevant WebViews are covered and that sensitive data is loaded into them. Ensure you keep a list of the sensitive data you expect to be cleaned up.
+5. Close the app.
+6. Use @MASTG-TECH-0002 to pull the contents of the `/data/data/<app_package>/app_webview/` directory or simply search for the sensitive data used in the WebView within that directory.
+
+## Observation
+
+The output should list the locations in the app where:
+
+1. The WebView enables particular storage areas.
```
Diolor: Or absence of such settings. It would mean defaults are used

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
+2. Cleanup APIs used during the current execution, or the lack of them.
+3. File system operations in the `/data/data/<app_package>/app_webview/` directory during app execution.
```
Diolor: How do we get those operations? Can we rewrite Step 6 and this to be better connected?

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
+4. The list of sensitive data expected to be cleaned up.
+5. The result of searching the contents of the `/data/data/<app_package>/app_webview/` directory for the sensitive data used in the WebView after closing the app.
+
+## Evaluation
+
+The test case fails if the app still has sensitive data on the `/data/data/<app_package>/app_webview/` directory after the app is closed. This could be due to the app not calling the relevant cleanup APIs after using the WebView.
+
+The test passes if all sensitive data used by the WebView is properly cleaned up using the relevant APIs, and no sensitive data remains in the `/data/data/<app_package>/app_webview/` directory after closing the app.
+
+!!! note
```
Diolor:
```suggestion
> note
```
?

tests-beta/android/MASVS-PLATFORM/MASTG-TEST-0320.md
```diff
@@ -0,0 +1,67 @@
...
+    It can be challenging to determine whether the right cleanup APIs were called for the enabled storage areas. @MASTG-KNOW-0018 describes the storage areas used by WebViews and the challenges of evaluating their cleanup.
+
+**Additional Guidance**:
+
+If you need more introspection during runtime, you can rerun the test with additional tracing of file system operations in the WebView storage directory.
+
+For example, you can use @MASTG-TECH-0033 to monitor file system operations in the `/data/data/<app_package>/app_webview/` directory. Regardless of whether the app uses these APIs directly, WebViews may use them internally when rendering content (e.g., JavaScript code using `localStorage`). So tracing calls to APIs such as `open`, `openat`, `opendir`, `unlinkat`, etc., can help identify file operations in the WebView storage directory.
```
Diolor: Just raising that this tends to be webview testing knowledge. Not how webviews work (KNOW) but how (hints) to test webviews. Could there be a better place for this ? KNOW, TECH?

demos/android/MASVS-PLATFORM/MASTG-DEMO-0082/MASTG-DEMO-0082.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Observation
+
+In the output we can see all instances of `deleteAllData()` of `WebStorage` called at runtime:
+
+{{ output.json }}
+
+We can also see that the output of the `adb shell grep` command shows no matches for the sensitive data used in the WebView:
+
+{{ output_adb_deletion.txt }}
```
Diolor: Not really important, does `output_adb_success.txt` / `output_adb_failure.txt` fit better?

PR #3584, opened 2025-12-12
Refactor base_script.js: Allow delay and prerequisites

Diolor:
### Improvements:
1. Allow delayed hooking
2. Allow change of classLoader (i.e. when the class in a different dex file)
3. Allow globs in class and method declaration
4. Allow prerequisite classes to be loaded before hooking the main class. This is not a "wait until" command, those classes will be loaded in memory explicitly.
5. Allow arrays or singled objects in prerequisites and methods
```js
const target = {
  delay: 1000, // optional delay
  hooks: [
    {
      class: "com.android.webview.chromium.*", // new: globs (* asterisks) can be added. Similar usage like frida-trace
      method: "deleteAllData",
      changeClassLoader: true,  // optional, change the classLoader of the class. See Frida's JS CLI: Java.classFactory 
      prerequisite: { // optional, can also be an array by setting a prerequisites (plural) value
        class: "android.webkit.WebStorage",
        method: "getInstance" // can also be an array by setting a methods (plural) value
      }
    }
  ]
}
``` 

### Housekeeping:
- Cluster similar functionality in methods 
- Convert some let to const
- Make output summary print `"overloads": []` instead of `overloads: [{"args": []}]` when args are empty.

---

For an actual example see #3558

### DIFF
--- a/.github/workflows/moderator.yml
+++ b/.github/workflows/moderator.yml
@@ -1,0 +1,163 @@
+name: AI Moderator
+
+on:
+  pull_request_target:
+    types: [opened]
+  issues:
+    types: [opened]
+
+permissions:
+  pull-requests: write
+  issues: write
+  models: read
+
+jobs:
+  quality-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Heuristic signals
+        id: heur
+        shell: bash
+        env:
+          TITLE: ${{ github.event.issue.title || github.event.pull_request.title }}
+          BODY: ${{ github.event.issue.body || github.event.pull_request.body }}
+        run: |
+          python - << 'PY'
+          import os, re, json
+          title = os.environ.get("TITLE") or ""
+          body = os.environ.get("BODY") or ""
+          text = (title + "\n" + body).strip()
+          urls = re.findall(r'https?://\S+', text)
+          repeated_urls = len(urls) != len(set(urls))
+          many_urls = len(urls) >= 3
+          empty_or_tiny = len(text) < 40
+          very_long = len(text) > 8000
+          keywordy = bool(re.search(r'\b(crypto|airdrop|giveaway|whatsapp|telegram|invest|forex)\b', text, re.I))
+          signals = {
+            "url_count": len(urls),
+            "repeated_urls": repeated_urls,
+            "many_urls": many_urls,
+            "empty_or_tiny": empty_or_tiny,
+            "very_long": very_long,
+            "keywordy": keywordy,
+          }
+          print("signals=" + json.dumps(signals, ensure_ascii=False))
+          PY
+          python - << 'PY'
+          import os, re, json
+          m = re.search(r'^signals=(.*)$', open("stdout.txt","r").read(), re.M) if False else None
+          PY
+        continue-on-error: true
+
+      - name: Detect spam or low-quality content
+        uses: actions/ai-inference@v1
+        id: ai
+        with:
+          model: openai/gpt-4o-mini
+          system-prompt: |
+            You are a conservative moderation assistant for a technical open source repository.
+            Use only the provided title and body and the provided heuristic signals.
+            Do not guess missing context.
+            Always return valid JSON with the exact keys requested.
+          prompt: |
+            Classify this GitHub ${{ github.event_name == 'issues' && 'issue' || 'pull request' }} as ok, needs-review, ai-generated, or spam.
+
+            Return a JSON object with exactly these keys.
+            verdict, confidence, reasons, evidence, action
+
+            Rules.
+            verdict is one of ok, needs-review, ai-generated, spam.
+            confidence is an integer from 0 to 100.
+            reasons is an array of 2 to 5 short strings describing why.
+            evidence is an array of 1 to 5 objects with keys quote and note, quote must be an exact substring from the title or body, or an empty string if no direct quote exists.
+            action is a short string describing what a maintainer should do next.
+
+            Heuristic signals, JSON.
+            ${{ steps.heur.outputs.signals }}
+
+            Title.
+            ${{ github.event.issue.title || github.event.pull_request.title }}
+
+            Body.
+            ${{ github.event.issue.body || github.event.pull_request.body }}
+
+      - name: Write audit summary
+        shell: bash
+        run: |
+          {
+            echo "### AI moderation report"
+            echo ""
+            echo "Raw model output."
+            echo '```json'
+            echo '${{ steps.ai.outputs.response }}'
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Apply label and comment if needed
+        if: steps.ai.outputs.response != ''
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const raw = `${{ steps.ai.outputs.response }}`.trim();
+            let report;
+            try {
+              report = JSON.parse(raw);
+            } catch (e) {
+              core.setFailed(`Model output was not valid JSON. Output was: ${raw}`);
+              return;
+            }
+
+            const verdict = String(report.verdict || '').trim();
+            const confidence = report.confidence;
+            const reasons = Array.isArray(report.reasons) ? report.reasons : [];
+            const evidence = Array.isArray(report.evidence) ? report.evidence : [];
+            const action = String(report.action || '').trim();
+
+            const number = ${{ github.event.issue.number || github.event.pull_request.number }};
+
+            const allowed = new Set(["ok", "needs-review", "ai-generated", "spam"]);
+            if (!allowed.has(verdict)) {
+              core.setFailed(`Unexpected verdict: ${verdict}`);
+              return;
+            }
+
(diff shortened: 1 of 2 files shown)


bernhste:
Very cool additions. I tested them it with the demos which use the base script and they work fine.

Just a few questions or remarks:

1. In general, I would avoid fix delays to solve problems because they can be inconstant. For example, if a class cannot be loaded yet, I would prefer a loop which checks again until a fix timeout is reached. Or do we have specific cases where hard timeouts are necessary?
2. I really like the `globs (* asterisks)`. We could also apply that to the methods in a future patch. An example with `SharedPreferences:`

```
var target = {
  category: "STORAGE",
  demo: "0059",
  hooks: [
    {
      class: "android.app.SharedPreferencesImpl$EditorImpl",
      methods: [
        "put*"
      ]
    },
 ...
```

3. I see the issue with different class loaders. But I think in this case, we should implicitly use the second method if the class cannot be loaded with the default loader. Or is there a use case, where it is beneficial if the user can determine which loader is used?
4. I started refactoring the frida.re decoders (#3612) because the code was a bit difficult to understand and maintain. I think decoupling the code for the class loading and the prerequisite parsing would help to make the code easier to understand and also testable.
5. Do you have demos in mind where we can use these new features? This would also be a good test case.

Diolor:
Hey @bernhste, thanks for testing and validating it works as expected!

As a start, check the example in [this PR, specifically the scripts.js file](https://github.com/OWASP/mastg/pull/3558/changes#diff-050374ac6c6e3108f5845c0e0a070535dc2d77d534b0b670a29f1adf8695ebe9R34). Under the hood, the `android.webkit.WebStorage` class is loaded by a different class loader (chromekit). So it was the PR that triggered the changes here. Merging this PR would enable that PR to use the generic script and a `hook.js` instead of the "manual" `script.js`.

1. We can ofc do that. Trying once looked stable for this first iteration, thus tried to avoid premature optimisation. I also looked at Frida internals to find "a prerequisite event" instead of a loop (such as "X class comes in memory" callback), but I don't think such a feature exists. If you know, that would be awesome!
2. For sure.
3. To make sure: do you suggest the following pattern: hook, catch the exception locally, change class loader, rehook? If we can identify it's a classloader exception, that could also work. Otherwise, we would always try twice before the script fails, making this slightly suboptimal. For now, with the explicit boolean, I transferred the responsibility to the "tester" to make it explicit, so there is no hidden logic, and we hook what we expect.
4. Awesome! Do you see the extraction as in scope for this PR, or would it make more sense to do it in a follow-up alongside other parts that might need extraction?

Also, please mark any other "must do"/"in scope" from your end. Otherwise, we could do those in follow-ups/create issues

bernhste:
When I had dynamically loaded classes/loaders, I always used a timeout and `Java.enumerateClassLoaders(callback)` or `Java.enumerateLoadedClasses(callback)` too.

But the demos must be deterministic because the build pipeline relies on them. So in case a class or class loader is dynamically loaded later during runtime (for example when pressing "Start" in the MAS Demo App), a fix timeout may sometimes fail unless the timeout is very large.

So if we have such a case, we could hook the constructor of `ClassLoader` and hook the method there? It's just a first idea, I haven't come across this scenario yet. As a first step this patch is a right step towards making the hooking more reliable. 

With regard to the modularization: I would to that in a later step. There is also another PR open (https://github.com/OWASP/mastg/pull/3557) which should be merged first before we change the base script further.  But I think the complexity at this point requires that we think about decoupling and testing soon. Of course Demos which use these features also help to find issues.

 "must do"/"in scope" should in my opinion be generally determined by the requirements of the Demos.

bernhste:
@Diolor Btw: I just needed the delay function myself today :) I intercepted `URI` and `URL` classes, and they are used heavily to the app at start time to load classes (`file://...dex`). There will be similar cases with file reader/writer.

So maybe it is just a general good idea to postpone the hooking a bit to make sure the app is properly initialized.

Diolor:
@bernhste, yes, anything UI-test-related could get tricky because device performance varies, which affects memory load times. :(

I would suggest "default disabling" such helpers (such as the delay) unless the majority of the test suite/devices would benefit from such default behavior. If we have only 2 cases so far, setting the default delay to 0 seems preferable. 
I have worked with only 1-2 hooks in this project; you might have a better sense of this need, of course.

We can also prioritise your loop/timeout suggestion.

utils/frida/android/base_script.js
```diff
@@ -205,91 +192,119 @@ function buildHookOperations(hook) {
...
-                "]. This hook will be skipped."
-            );
-            errors.push("Overload not found for " + hook.class + ":" + hook.method + " with args [" + argsExplicit.join(", ") + "]");
-          }
-        }
-      } catch (e) {
-        let errMsg = "Failed to process method '" + hook.method + "' in class '" + hook.class + "': " + e;
-        console.warn("Warning: " + errMsg);
-        errors.push(errMsg);
+  function loadPrerequisite(prerequisite) {
```
bernhste: I would not use plurals for multiple prerequisites but rather just an array to avoid an inconsistent API. The `hook` object would then look like that:

```
{
  class: "com.android.webview.chromium.*",
  method: "deleteAllData",
  changeClassLoader: true,
  prerequisite: [
    {
      class: "android.webkit.WebStorage",
      method: ["getInstance"]
    }
  ]
}
```

Later we want to migrate it to TypeScript, and then a simple array is better.

This would also simplify `resolveClass()` because we don't have to treat single and multiple prerequisites differently.

By the way, I noticed that method(s) is the same. I think we should change that there too (but in another PR).

utils/frida/android/base_script.js
```diff
@@ -205,91 +192,119 @@ function buildHookOperations(hook) {
...
   let errors = [];
 
-  try {
-    // Invalid configuration: methods + overloads (logged elsewhere)
-    if (hook.methods && hook.overloads && hook.overloads.length > 0) {
-      let errInvalid = "Invalid hook configuration for " + hook.class + ": 'overloads' is only supported with a singular 'method', not with 'methods'.";
-      console.error(errInvalid);
-      errors.push(errInvalid);
-      return {operations: operations, count: 0, errors: errors, errorCount: errors.length};
+  function loadClass(clazz, method) {
```
bernhste: The name of this function can indicate that all classes are loaded with it, but it calls the prerequisite method without parameters. What about renaming it to `callPrerequisiteMethod()`. 

But I think we should refactor the way classes and methods are loaded after the migration to the dedicated project.

utils/frida/android/base_script.js
```diff
@@ -205,91 +192,119 @@ function buildHookOperations(hook) {
...
+      prerequisite.methods.forEach(method => {
+        loadClass(prerequisite.class, method)
+      })
+    }
+    if (prerequisite.method) {
+      loadClass(prerequisite.class, prerequisite.method)
+    }
+  }
+
+  function resolveClass(inputClass, method) {
```
bernhste, line 380: I tested the wildcard class lookup with:

```
const target = {
  delay: 1000,
  hooks: [
    {
      class: "com.*",
      method: "$init",
    }
  ]
}
```

Here I would expect that all constructors of all classes starting with `com.` would be hooked. But only the first match was hooked:

```
[Android Emulator 5554::org.owasp.mastestapp ]-> {
  "type": "summary",
  "hooks": [
    {
      "class": "com.sun.security.cert.internal.x509.X509V1CertImpl",
      "method": "$init",
      "overloads": [
        {
          "args": [
            "java.io.InputStream"
          ]
        },
        {
          "args": [
            "[B"
          ]
        }
      ]
    }
  ],
  "totalHooks": 3,
  "errors": [],
  "totalErrors": 0
}
```

If we implement a wildcard matching, we should set some limit on how many classes and methods can be looked up. Because the expected scope of this example would be huge.

bernhste (changes requested):
I tested the new functionality and added some comments. The only feature I did not test yet was the `changeClassLoader`. But I understand how it works. 

Do you have some examples that use `changeClassLoader`? Because this would be interesting for testing.

The question now is when/how we merge.

utils/frida/android/base_script.js
Diolor: As you said in the end, I followed the existing pattern based on hook.method/hook.methods.

What issue do you think will come up with a TypeScript migration? I also think the added complexity is not that much (3 lines). Regardless, I can drop the singular cases; however, in that case, the keys should be marked in plural when we have an array.

```
{
  class: "com.android.webview.chromium.*",
  methods: ["deleteAllData"],
  changeClassLoader: true,
  prerequisites: [
    {
      class: "android.webkit.WebStorage",
      methods: ["getInstance"]
    }
  ]
}
```

bernhste:
@Diolor The frida.re script moved to https://github.com/cpholguera/frooky.

I suggest that we manually apply these changes to the code base there since there are some changes done to the code base. I would open a new issue there so we can close this one.

Is that ok for you?

bernhste:
Perfect. So lets move this discussion to https://github.com/cpholguera/frooky/issues/6

PR #3612, opened 2026-01-06
Decouple frida.re script (now uses frida-compiler)

### LINKED ISSUE #3451: Improve Frida util: Use frida-compiler for code generation
At the moment, the JavaScript used to intercept API is relatively simple. The demos use a `run.sh` script which concatenates the `hook.js` into the base script. If we want to add additional functionalities as described in https://github.com/OWASP/mastg/issues/3365 we should use `frida-compiler` so we can separate the different parts of the JS code. For example the value decoders should be separated form the hook-code for better maintainability.

And while the codebase is still small I recommend migrating it to TypeScript.

bernhste:
This PR closes #3451

## Description

The for the demos, the script was compiled using string concatenation. Also, all decoders were implemented in one file (`decoders.js`). 

I refactored the code base to use `frida-compiler`. Now the decoders are decoupled and easier to maintain. They are all in the folder `./utils/frida/android/decoders`. One module for each type, a util module and a module which assigns the decoders to the type.

The base script always only used `decodeArguments()`. This does not change.

I tested the refactored code with all demos which use this frida.re script so far (MASTG-DEMO-0058, MASTG-DEMO-0059, MASTG-DEMO-0060, MASTG-DEMO-0081). They all produce the same output.

The base script itself did change slightly (now using modules). Other than that, no changes. So this should not conflict with #3584.

This should make it much easier maintaining and adding decoders and tests, which I would like to implement.

-------------------

### AI Tool Disclosure

- [x] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `kagi assistant`
    - LLMs and versions: `claude sonnet 4.5`
    - Prompts: `take this code <code> and decouple modules for each decoder type (all the cases in the big switch case). make a file for each type.  name the files so that it is clear what type it decodes (example: java.io.File.js) name the decode function accordingly (example decodeJavaSecurityPrivateKey()) ONLY export decodeArguments() some of the functions are not type decoder, collect them in a util module. don't change anything at the code nor the comments they should remain as they are fully.`

-------------------

[x] I have read the contributing guidelines.

### DIFF
--- a/utils/frida/android/run.sh
+++ b/utils/frida/android/run.sh
@@ -9,7 +9,6 @@ randomNumber=$RANDOM
 workDir=/tmp/frida_$randomNumber
 mkdir $workDir
 
-
 # copy relevant files temporary work folder
 cp $hookPath $workDir
 cp -rf ../../../../utils/frida/android/* $workDir


utils/frida/android/run.sh
```diff
@@ -1,22 +1,24 @@
...
 
 hookPath=$1
-hook=$(cat "$hookPath")
-decoderScript=$(cat "$(dirname $0)"/android_decoder.js)
+demoPath=$(pwd)
 fridaScript=$(cat "$(dirname $0)"/base_script.js)
+
+# create temporary work folder
 randomNumber=$RANDOM
+workDir=/tmp/frida_$randomNumber
```
Diolor, line 9: Potentially out of scope, but let's better use a project-local temporary folder and not a computer global one?
Plus add it on.gitignore

I could do it in #3584 but then we would conflict

utils/frida/android/decoders/android.database.Cursor.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Decodes a `android.database.Cursor` object.
+ * @param {object} value - Reference to the object.
+ * @returns {string} The decoded rows and columns.
+ */
+function decodeAndroidDatabaseCursor(value) {
+  let out = "";
```
Diolor, line 7: some `let`s could be `const`s, + other files

bernhste:
will reopen after migration of frida tool to dedicated project.

PR #3619, opened 2026-01-10
Use project-local .tmp directory for frida scripts

prasunsrivastav123-lang:
- Use a project-local temporary directory instead of /tmp
- Add .tmp/ to .gitignore
- Avoid global temp files and improve safety

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ site/
 .cache/
 references.json
 .obsidian/
+.tmp/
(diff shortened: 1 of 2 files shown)


cpholguera:
This topic is already being addressed on another PR so we're going to close this PR. Thanks @prasunsrivastav123-lang and please always prefer to pick task that already have an issue or propose one yourself before opening a PR. See https://mas.owasp.org/contributing/1_How_Can_You_Contribute/

PR #3642, opened 2026-01-21
Add event name to allowlist gate in workflow

cpholguera:
This pull request updates the workflow logic in `.github/workflows/moderator.yml` to improve how user permissions are checked for different GitHub event types. The main change is that the workflow now determines the relevant user (the "subject") based on the event type (pull request, issue, or other), rather than always using the actor who triggered the event.

### DIFF
--- a/.github/workflows/moderator.yml
+++ b/.github/workflows/moderator.yml
@@ -31,7 +31,7 @@ jobs:
           pr_author = (os.environ.get("PR_AUTHOR") or "").strip().lower()
           issue_author = (os.environ.get("ISSUE_AUTHOR") or "").strip().lower()
       
-          allow = {"cpholguera", "sushi2k"}
+          allow = {"cpholguera", "sushi2k", "Copilot"}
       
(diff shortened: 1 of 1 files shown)


.github/workflows/moderator.yml
```diff
@@ -18,14 +18,33 @@ jobs:
...
           python - << 'PY'
           import os
+      
+          event = (os.environ.get("EVENT_NAME") or "").strip()
           actor = (os.environ.get("ACTOR") or "").strip().lower()
-          allow = {"cpholguera", "sushi2k", "Copilot"}
+          pr_author = (os.environ.get("PR_AUTHOR") or "").strip().lower()
+          issue_author = (os.environ.get("ISSUE_AUTHOR") or "").strip().lower()
+      
+          allow = {"cpholguera", "sushi2k"}
```
cpholguera:
```suggestion
          allow = {"cpholguera", "sushi2k", "Copilot"}
```

PR #3814, opened 2026-05-22
Fix ios path

Diolor:
Fix ios path


.github/workflows/build-android-demos.yml
```diff
@@ -10,6 +10,7 @@ on:
   pull_request:
     branches:
       - master
+      - Diolor:fix-ios-path
```
Diolor: remove before merge

.github/workflows/build-ios-demos.yml
```diff
@@ -10,6 +10,7 @@ on:
   pull_request:
     branches:
       - master
+      - Diolor:fix-ios-path
```
Diolor: remove before merge

PR #3881, opened 2026-06-15
Enhancements after PR #3807 for TESTs 0372-0375 and DEMOs 0136-141

jacobocasado:
Review branch for OWASP/mastg#3807.


jacobocasado:
Closing in favor of #3883, which targets current master after #3807 was merged and uses the real published IDs.

PR #3883, opened 2026-06-15
Review implicit intents content

jacobocasado:
Follow-up review for OWASP/mastg#3807


knowledge/android/MASVS-CODE/MASTG-KNOW-0138.md
cpholguera, line 1: If we remove 137 we'll need to renumber this one to 137.

TheDauntless (commented):
Great PR! Made a few small modifications but no further action is needed.

Also created https://github.com/OWASP/mastg/issues/3906 as a result

TheDauntless (approved):

PR #3930, opened 2026-07-26
Enhancement of OAuth 2.0 Best Practices section

### LINKED ISSUE #1174: update oauth2
Enhance Oauth 2 recommendations based on evaluationing the following RFC: https://tools.ietf.org/html/rfc8252

mateofumis:
This PR closes #1174

## Description

This PR contains the updates of the proposed "**Option B**", enhancing the OAuth 2.0 [Best Practices](https://mas.owasp.org/MASTG/0x04e-Testing-Authentication-and-Session-Management/#best-practices) section on the chapter [0x04e-Testing-Authentication-and-Session-Management](https://mas.owasp.org/MASTG/0x04e-Testing-Authentication-and-Session-Management).

---

## AI Tool Disclosure

Check exactly one option.

- [x] This contribution does not include AI-generated content.
- [ ] This contribution includes AI-generated content.

---

## Contributor Checklist

- [x] I have read and understood the contributing guidelines.
- [x] I followed the project style guide.
- [x] I validated the technical correctness of my changes and understand the topic.
- [x] This PR adds clear value and is not spam or low-effort content.

### DIFF
--- a/Document/0x04e-Testing-Authentication-and-Session-Management.md
+++ b/Document/0x04e-Testing-Authentication-and-Session-Management.md
@@ -239,7 +239,11 @@ Some of the best practices include but are not limited to:
 
 When implementing OAuth2 in **native apps** that need to interact with social media accounts, the use of embedded user agents such as WebView or WKWebView must always be avoided, because the host app has full control over the embedded view and can record user inputs such as usernames and passwords.
 
-Instead of using _custom schemes_, for example `myapp://callback`, App Links (Android) or Universal Links (iOS) must always be used. If a second app registers the same custom scheme as the legitimate, it could lead the user to a hijack scenario when the malicious app steals the secret code.
+Instead, a safer implementation is the use of _external user agents_, such as the system browser or in-app browser tabs, because they run in a separate instance and prevent the host app from capturing the user's credentials.
+
+Claimed HTTPS redirect URIs, such as App Links (Android) or Universal Links (iOS), should be used where possible. A valid alternative is the implementation of _private-use URI scheme_, for example `com.example.app:/oauth2redirect/example-provider`.
+
+However, if a second app registers the same custom scheme as the legitimate, it could lead the user to a hijack scenario where the malicious app steals the authorization code.
 
 ## User Logout
 
--- a/Document/0x04e-Testing-Authentication-and-Session-Management.md
+++ b/Document/0x04e-Testing-Authentication-and-Session-Management.md
@@ -237,6 +237,10 @@ Some of the best practices include but are not limited to:
     - Remember that an attacker who has stolen tokens can access their scope and all resources associated with them if the app uses access tokens as bearer tokens with no other way to identify the client.
     - Store refresh tokens in secure local storage; they are long-term credentials.
 
+When implementing OAuth2 in **native apps** that need to interact with social media accounts, the use of embedded user agents such as WebView or WKWebView must always be avoided, because the host app has full control over the embedded view and can record user inputs such as usernames and passwords.
+
+Instead of using _custom schemes_, for example `myapp://callback`, App Links (Android) or Universal Links (iOS) must always be used. If a second app registers the same custom scheme as the legitimate, it could lead the user to a hijack scenario when the malicious app steals the secret code.
+
 ## User Logout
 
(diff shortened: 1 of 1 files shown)


bernhste:
Hey. Thanks for the input. 

Here the malicious actor is the app developer trying to steal the credentials of the user. 

The user must further be aware about the technical details in order to detect a possible attack: If no external/system browser, with a potentially already established session, is opened, the app developer could steal the password. But in the end, the user has to trust the app developer. 

I'm thinking about how to align the risk with the [testing profiles](https://mas.owasp.org/MASTG/0x03b-Testing-Profiles/#the-mas-testing-profiles). In general, they do not assume that the app developer is malicious. And if the app developer were malicious, they would not implement this recommendation.

But since it is generally a better design to use the system browser, I think we should mention it somewhere in MAS. Maybe as a [best practice](https://mas.owasp.org/MASTG/best-practices/#). 

Or one could also argue, that it is related to [MAS-P](https://mas.owasp.org/MASTG/0x03b-Testing-Profiles/#mas-p-baseline-privacy-profile), because it protects the users credentials from being stolen by the app developer. In this case, we could also make a test / demo.

cpholguera:
In these additions, you mention what **must not** be used, but the RFC provides alternatives that we should mention here. After this PR ensures everything is captured, we should revisit this topic and create best practice articles, as well as tests and demos, as Stefan suggests. For now, let's finish this PR so we can close the issue and create new ones for the rest of the work.

cpholguera:
@bernhste let's discuss the "malicious developer" topic and its relationship with profiles on our next MAS Task Force sync

mateofumis:
@cpholguera Thanks. I'll take another look at the **RFC** regarding the alternatives that should be mention in this PR, as you said, and then apply those changes properly in a new commit.

Document/0x04e-Testing-Authentication-and-Session-Management.md
```diff
@@ -237,6 +237,14 @@ Some of the best practices include but are not limited to:
     - Remember that an attacker who has stolen tokens can access their scope and all resources associated with them if the app uses access tokens as bearer tokens with no other way to identify the client.
     - Store refresh tokens in secure local storage; they are long-term credentials.
 
+When implementing OAuth2 in **native apps** that need to interact with social media accounts, the use of embedded user agents such as WebView or WKWebView must always be avoided, because the host app has full control over the embedded view and can record user inputs such as usernames and passwords.
+
+Instead, a safer implementation is the use of _external user agents_, such as the system browser or in-app browser tabs, because they run in a separate instance and prevent the host app from capturing the user's credentials.
+
+Claimed HTTPS redirect URIs, such as App Links (Android) or Universal Links (iOS), should be used where possible. A valid alternative is the implementation of _private-use URI scheme_, for example `com.example.app:/oauth2redirect/example-provider`.
```
cpholguera: Why would we list this as a "valid alternative" when it's not the best practice?

```suggestion
Even though RFC 8252 permits the use of private-use URI schemes, they should be avoided where possible because another app can register the same custom scheme and intercept the authorization code.

Instead, claimed "https" scheme URIs, such as Android App Links or iOS Universal Links, should be used where possible, as they allow the operating system to associate the redirect URI with the legitimate app.
```

Document/0x04e-Testing-Authentication-and-Session-Management.md
```diff
@@ -237,6 +237,14 @@ Some of the best practices include but are not limited to:
...
+
+However, if a second app registers the same custom scheme as the legitimate, it could lead the user to a hijack scenario where the malicious app steals the authorization code.
```
cpholguera: Why would we list this as a "valid alternative" when it's not the best practice?

```suggestion
Even though RFC 8252 permits the use of private-use URI schemes, they should be avoided where possible because another app can register the same custom scheme and intercept the authorization code.

Instead, claimed "https" scheme URIs, such as Android App Links or iOS Universal Links, should be used where possible, as they allow the operating system to associate the redirect URI with the legitimate app.
```

cpholguera (changes requested):

Document/0x04e-Testing-Authentication-and-Session-Management.md
mateofumis: Makes sense. I added that based on how I understood your earlier comment about mentioning the RFC alternatives here.

cpholguera (approved):

